Issue #3, opened 2022-08-04
Evaluate implementation of an SCM class


robertness:
@adam2392 comments "I think we should add a class for SCMs since causal graphs really stem from SCMs...".

This task is not implementation specific, we just need a path forward.  For example, perhaps we just agree with the CGM developers to move [those abstractions](https://github.com/py-why/dowhy/blob/77f7064c85b3de740b5beefb0ad7067ffdb024fd/dowhy/gcm/graph.py#L93) here?

PR #11, opened 2022-08-30
[DOC] Adding basic conceptual example of causal graphs

adam2392:
I've been merging in some PRs without reviews for the sake of getting upstream PRs to work. Now I think we're at a stage where we can improve the codebase, but also not be a bottleneck for super common algorithms upstream (e.g. FCI). This PR is now a reviewable PR for the sake of converging on the communication of concepts to end-users.

Addresses partially: #3 

- Based on the previous PRs: #8 , #9 , #10 this example now walks through the concepts of fundamental graphs that are used in constraint-based structure learning and causal inference

## Before submitting

## After submitting


adam2392:
Here is the example rendered: https://output.circle-artifacts.com/output/job/ea0363b2-a2fb-420d-8834-fada0ce23cf4/artifacts/0/dev/auto_examples/intro_causal_graphs.html#sphx-glr-auto-examples-intro-causal-graphs-py

pyproject.toml
```diff
@@ -69,11 +69,13 @@ sphinx-copybutton = { version = "^0.5.0" }
...
-dowhy = { version = "^0.8" }
 ipython = { version = "^7.4.0" }
 nbsphinx = { version = "^0.8" }
-scikit-learn = { version = "^1.1.0" }
-pandas = { version = "^1.1" }
+dowhy = { version = "^0.8" }
+scikit-learn = { version = "^1.1.0" }  # needed in dowhy's package
+pandas = { version = "^1.1" }  # needed in dowhy's package
+joblib = { version = "^1.1.0" }  # needed in dowhy's package
+tqdm = { version = "^4.64.0" }  # needed in dowhy's package
```
petergtz, line 78: This should get automatically pulled in through dowhy. Unless something is wrong (and you're not directly using these package in pywhy-graphs), this should not be necessary.

examples/intro_causal_graphs.py
```diff
@@ -0,0 +1,235 @@
...
+
+# Now there is a bidirected edge between 'x' and 'y'
+draw(admg)
+
+# Now if one queries the parents of 'y', it will not show 'xy' anymore
+print(list(admg.predecessors("y")))
+
+# The bidirected edges also form a cluster in what is known as "confounded-components", or
+# c-components for short.
+print(f"The ADMG has c-components: {list(admg.c_components())}")
```
petergtz, line 166: Is "c_components" a well-known expression in the science community or would this benefit from the verbose name "confounded_components".

examples/intro_causal_graphs.py
```diff
@@ -0,0 +1,251 @@
...
+====================================================
+
+Causal graphs are graphical objects that attach causal notions to each edge
+and missing edge. We will review some of the fundamental causal graphs used
+in causal inference, and their differences from traditional graphs.
+"""
+
+# Authors: Adam Li <adam2392@gmail.com>
+#
+# License: BSD (3-clause)
```
petergtz, line 15: Let's keep these out. Authors are in git commits and the license has a dedicated file (which avoids that we get conflicting licenses in here).

pywhy_graphs/algorithms/pag.py
```diff
@@ -250,7 +251,7 @@ def discriminating_path(
         # check distance criterion to prevent checking very long paths
         distance += 1
         if distance > 0 and distance > max_path_length:
-            logger.warn(
+            warn(
```
petergtz, line 254: I think I would not change this. See [here](https://stackoverflow.com/a/14762106), why:
> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
>
> logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

I think the latter applies here, no?

petergtz (commented):
Hey @adam2392, I've done a very superficial review (given my lack of theoretical background in the science topic) and would expect others to weigh in too.

A general comment: to make these kinds of reviews easier for reviewers, please don't mix different changes into one PR. I think the changes in `pag.py` and `admg.py` should not be in this PR (and probably those in `pyproject.toml` neither).

I know that it's often annoying to fiddle around with different branches, separating different changes into those different branches, etc. But there's one huge benefit I've observed in the past: reviews and especially approvals will happen faster, because there's less to review, **and** one requested change does not block all the other changes you made. So I'd encourage you to try it out.

examples/intro_causal_graphs.py
adam2392: I think the string can be more explicit, but in general anyone working in causality from graphical model perspective would get exposed to c-components (imo), but open to changing the property name too.

pyproject.toml
adam2392: Trying this now...

adam2392:
> I know that it's often annoying to fiddle around with different branches, separating different changes into those different branches, etc. But there's one huge benefit I've observed in the past: reviews and especially approvals will happen faster, because there's less to review, **and** one requested change does not block all the other changes you made. So I'd encourage you to try it out.

Hey @petergtz sure I can do that. I figured small changes on top of the main contributing example were fine, but I guess they kind of stacked up to a significant # LOC :/.

I'm okay with waiting on this one to merge in. This is an example really trying to communicate core concepts, and I think in general will be one of the first things a user might look at, so def a higher bar. 

I'll try to be more cognizant of breaking up the proposed changes more.

robertness (approved):

PR #13, opened 2022-09-05
Add MVG projects docs

petergtz:
Hi @adam2392,

see https://github.com/py-why/dodiscover/pull/36 for what this is about.

I listed only you for now as maintainer. All up for discussion of course :-).


adam2392 (approved):

adam2392:
Sure sounds good! It would be great to get some more maintainers tho and ppl invested, so we can raise it to newcomers as well :)

Issue #57, opened 2023-02-20
Conversion API to other causal libraries' graph


adam2392:
**Is your feature request related to a problem? Please describe.**
We should expose an API (there is one untested really rn) for exporting to the following packages:

- tigramite (3D time-series graph that supports time-series DAG, ADMG, PAG) -> WIP
- pgmpy (DAG and PDAG)
- causal-learn (currently we have this implemented) -> done
- dagitty (DAG, ADMG) -> handled by pygraphviz
- pcalg -> done
- networkx to a DiGraph with a specific structure
- numpy (tigramite, causal-learn, pcalg are all numpy-like array structures) -> done
- ananke -> @jaron-lee to handle
- tetrad text -> done
- bnlearn -> handled by pygraphviz I think

This will enable pywhy-graphs to be interoperable across many other packages and also allow other users to take their graphs and convert to pywhy-graphs for the sake of using it in `dodiscover` and `dowhy`.

**Describe the solution you'd like**
If the other package is not in R, we can directly test roundtrip compatibility.

If the other package is in R, we should just do some basic testing that the format is "as expected". I think loading in R runtime is a whole other beast we don't want to figure out how to do in the CI pipeline.

The solution requires the following for each export:

io roundtrip: That is the basic graph structure should be preserved. Some metadata might be lost depending on the format each package stores graphs in. When possible, metadata should be preserved. A unit test would look like the following where the edge and node structure is preserved.

```Python
pcalg_G = pywhy_graphs.export(G, 'pcalg')
G_copy = pywhy_graphs.import(G, 'pcalg')

for edge_type in G.edge_types:
    assert nx.is_isomorphic(G.get_graphs(edge_type), G_copy.get_graphs(edge_type))
```

**Additional context**
n/a

jaron-lee:
I would also be happy to extend the API to ananke!

adam2392:
I think based on today's discussion, this seems pretty high value. So I'ma start chipping away at this. Happy to review PR related to ananke. 

Re all these packages btw, we'll only add them as soft dependencies for unit-testing purposes. This will go into the poetry group `test` in the `pyproject.toml` file

jaron-lee:
Will get started on this later this week, on the ananke side I need to remove a hard dependency on `graphviz` and make sure that didn't break anything.

adam2392:
[...]

PR #60, opened 2023-02-27
[ENH] Export and import graph classes from other common interfaces

adam2392:
Towards #57 

- numpy export
- pcalg export
- tetrad export
- causal learn export now tested
- I removed "integration tests", since pywhy-graphs is more of an upstream package and we want to integrate support for exporting/importing other graph packages

Now need to add manual tests using what is present in: https://github.com/py-why/causal-learn/blob/main/tests/fci-test-data/result-py-FCI-8.txt for tetrad.

## Before submitting

## After submitting


jaron-lee:
At some point it would be great if you could pick one of `causallearn.py, tests/test_causallearn.py` or `causal_learn.py, tests/test_causal_learn.py`, it's ruining my Vim projectionist heuristics when I switch back and forth 😂

jaron-lee:
So I notice that `tigramite` has a hard dependency on exactly one patch version of scipy (`1.8.0`) - so happens that ananke requires `^1.8.1`. In this particular case I could go and decrement the patch version in ananke, but is there another way to set this up in poetry? The version conflict here is not that bad given that we don't actually need both of these packages at the same time per se.

jaron-lee:
Ran into another issue - when I try and run unit tests python is picking up `networkx.py` as the `networkx` package (instead of whatever comes from `pip`). Is there a way to distinguish between the local module vs the pip package?

adam2392:
What's the issue? The export submodule or the networkx submodule for mixed edge graphs?

adam2392:
> So I notice that `tigramite` has a hard dependency on exactly one patch version of scipy (`1.8.0`) - so happens that ananke requires `^1.8.1`. In this particular case I could go and decrement the patch version in ananke, but is there another way to set this up in poetry? The version conflict here is not that bad given that we don't actually need both of these packages at the same time per se.

Ah yeah that's an issue. If it's easy to do it for ananke, then that would be great! 

I'll also submit an issue to tigramite to make the dependency not hard. I know the maintainer.

jaron-lee:
> What's the issue? The export submodule or the networkx submodule for mixed edge graphs?

I think the export submodule - basically running tests from that folder would get python confused about which networkx to import. Might just be a weird issue on my test runner's end tho.

adam2392:
Ah I see. We can def rename things if it's an issue. 

You can just run pytest from the root of the package folder. E.g.

Pytest ./pywhy_graphs/...

And it should work.

adam2392:
This is somewhat of a big PR, and happy to break it up, but just wanted to get a high level review of the code from you @jaron-lee and @robertness to make sure the API looks good to you?

pywhy_graphs/export/tetrad.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Notes
+    -----
+    See tetrad documentation: https://cmu-phil.github.io/tetrad/manual/.
+
+    Warning: There is no guarantee or check that your file adheres to the
+    tetrad format.
+    """
+    # instantiate the type of causal graph
+    if graph_type == "dag":
```
jaron-lee, line 31: I think I have seen different ways of dealing with DAGs - should we use nx.DiGraph or send it to pywhy_graphs? ADMG here seems like a hack. Maybe an in-house pywhy_graphs DAG class is needed?

pywhy_graphs/testing.py
```diff
@@ -0,0 +1,27 @@
+import networkx as nx
+
+
+def assert_mixed_edge_graphs_isomorphic(G1, G2):
```
jaron-lee, line 4: This looks good, I presume all roundtrip tests should use this testing function ?

pywhy_graphs/export/tests/test_tetrad.py
```diff
@@ -0,0 +1,19 @@
...
+import pywhy_graphs
+from pywhy_graphs.export import graph_to_tetrad, tetrad_to_graph
+from pywhy_graphs.testing import assert_mixed_edge_graphs_isomorphic
+
+root = Path(pywhy_graphs.__file__).parents[1]
+filename = root / "pywhy_graphs" / "export" / "tests" / "test_pag_tetrad.txt"
+
+
+def test_to_tetrad(tmp_path):
+    G = tetrad_to_graph(filename, graph_type="pag")
```
jaron-lee, line 12: What happens if a user has `graph_type="dag"` instead here? Will this function detect that there are `o->` edges and prevent that? Clearly the imported file does not represent a DAG so we shouldn't let the user make this mistake. Or we should have clear tests to show what happens when you import a graph under a different graph assumption than the one it was created with

jaron-lee:
> > In the `<package>_to_graph` function we have this argument `graph_type`. Should this be a mandatory argument? Sometimes (e.g. in Ananke) the graph already carries a type which has a corresponding equivalent in pywhy_graphs, so `graph_type` can be easily inferred.
> 
> The `graph_type` parameter is just used to make sure the user explicitly spells out which graph to convert to because sometimes it is not clear. For example, pcalg encodes CPDAG and PAG as a numpy array, but there 1s and 0s mean different things, so we need to have different logic. If ananke can let us know easily which graph type to use, then I would say the `graph_type` parameter is not needed.
> 
> Note: not all graph types are supported in each export function. E.g. pcalg. We can change that tho, but I avoided the extra work.
> 
> > Additionally the `graph_to_<package>` function is typed for `MixedEdgeGraph` input. Do we expect users to actually pass in `MixedEdgeGraph`, or will we expect them to pass a child class like `ADMG`? Or should we expect both? This also has implications for exporting to Ananke.
> 
> `graph_to_<package>` can usually eat any graph because we have semantic separation of all types of graphs. The ADMG subclasses a MixedEdgeGraph but with very specific structure.
> 
> Lmk how that sounds. If there's anything you want me to change for downstream ananke PR, lmk!

This all looks good. What I'll do is if a user passes in something of type `MixedEdgeGraph` then I'll just handle it equivalently on the Ananke side (`Graph`) without doing any class inferring.

jaron-lee (approved):

pywhy_graphs/export/tetrad.py
adam2392: True. Perhaps `nx.DiGraph` with some added semantics to ensure construction is acyclic... But it'd be too expensive to check acyclicity every time someone adds a node. 

On one hand it seems overkill to recreate a DAG class when nx.DiGraph works. On the other hand, the mental switching is a bit weird. 

Tbh not sure exactly which way to go.

pywhy_graphs/testing.py
adam2392: Yep!

FYI I realized there's a lot of LOC scattered in the codebase doing this.

pywhy_graphs/export/tests/test_tetrad.py
adam2392: I'm going to leave this for another PR.

PR #63, opened 2023-03-09
Add ananke graph export functionality

### LINKED ISSUE #57: Conversion API to other causal libraries' graph

jaron-lee:
Fixes #57 for the `ananke-causal` library

- Implements an export and import API for `ananke-causal` graphs

## Before submitting

- [x] I've read and followed all steps in the [Making a pull request](https://github.com/py-why/pywhy-graphs/blob/main/CONTRIBUTING.md#making-a-pull-request)

## After submitting


pyproject.toml
```diff
@@ -50,7 +50,8 @@ pytest = "^7.1.2"
 pytest-cov = "^3.0.0"
 memory_profiler = { version = "^0.60.0" }
 causal-learn = { version = "^0.1.2.8" }
-tigramite = { version = "^5.1" }
+#tigramite = { version = "^5.1" }
```
adam2392, line 53: Let's just remove tigramite for now.

adam2392: It's not used.

jaron-lee:
Did we reach a consensus of whether we should have a pywhy_nx.DAG class? I feel like while it would duplicate some nx functionality, for the sake of consistency it would be nice to have. 

A related issue is that in Ananke we support all combinations of directed, bidirected, and undirected graphs. However, while we represent undirected and bidirected graphs through an nx.Graph, these don't behave the same way in downstream algorithms (e.g. m-separation works differently in these graphs) so we would need to wrap these in MixedEdgeGraphs. Down the line it would be nice to provide classes directly for some of these graphs rather than calling MixedEdgeGraph directly.

adam2392:
> Did we reach a consensus of whether we should have a pywhy_nx.DAG class? I feel like while it would duplicate some nx functionality, for the sake of consistency it would be nice to have. 

For now I'd say let's just use networkx until we integrate pywhy graphs with some downstream functionality. 

> A related issue is that in Ananke we support all combinations of directed, bidirected, and undirected graphs. However, while we represent undirected and bidirected graphs through an nx.Graph, these don't behave the same way in downstream algorithms (e.g. m-separation works differently in these graphs) so we would need to wrap these in MixedEdgeGraphs. Down the line it would be nice to provide classes directly for some of these graphs rather than calling MixedEdgeGraph directly.

WDYM here? Can you clarify?

The purpose of export is to convert other graph packages to ours and vice versa. When in ours they should work nicely with all the algorithms.

jaron-lee:
> WDYM here? Can you clarify?
> 
> The purpose of export is to convert other graph packages to ours and vice versa. When in ours they should work nicely with all the algorithms.

Yea all I'm saying is that for certain graph classes like `ADMG` or `PAG` we have pywhy_graph classes for them, but e.g chain graphs (directed + undirected) currently lack a corresponding class. So I'm forced to call MixedEdgeGraph directly. At some point we should change that, although for now I think it's fine

jaron-lee:
> For now I'd say let's just use networkx until we integrate pywhy graphs with some downstream functionality.

OK - but e.g. in causallearn.py it looks like we are using pywhy_nx.ADMG for DAGs vs nx.DiGraph.

adam2392:
> OK - but e.g. in causallearn.py it looks like we are using pywhy_nx.ADMG for DAGs vs nx.DiGraph. Should we patch that up?

Oh yeah. I think probably a good call to convert that to a nx.DiGraph for now. The reason I'm inclined to hold off is I feel like once we "use the pywhy-graphs API" more, we'll get a sense of the different shortcomings, so don't want to commit to adding a whole new class too early.

adam2392:
> Yea all I'm saying is that for certain graph classes like `ADMG` or `PAG` we have pywhy_graph classes for them, but e.g chain graphs (directed + undirected) currently lack a corresponding class. So I'm forced to call MixedEdgeGraph directly. At some point we should change that, although for now I think it's fine

Ah I see. Yeah we don't have a chain graph, so maybe for now don't support the exporting. 

Can you educate me a bit and specify what's the difference between a chaingraph and ADMG/PAG/CPDAG?

jaron-lee:
> > Yea all I'm saying is that for certain graph classes like `ADMG` or `PAG` we have pywhy_graph classes for them, but e.g chain graphs (directed + undirected) currently lack a corresponding class. So I'm forced to call MixedEdgeGraph directly. At some point we should change that, although for now I think it's fine
> 
> Ah I see. Yeah we don't have a chain graph, so maybe for now don't support the exporting.
> 
> Can you educate me a bit and specify what's the difference between a chaingraph and ADMG/PAG/CPDAG?

Yeah so a chain graph just has directed and undirected edges. Directed edges mean what they usually mean in a DAG, and the undirected edges represent an equilibrium process between two variables. There's a condition that you can't have a partially directed cycle (must not be able to orient an undirected edge to form a directed cycle). If that is satisfied, then this graph has a factorization that basically looks like the factorization of a DAG and a Markov random field squashed together (first you factorize according to the directed edges, which gives you blocks of undirected connected components, and then you factorize each block as you would an MRF).

These actually do come up from time to time in causality so I think it might be worth adding eventually.

I ended up implementing the export functionality directly from MixedEdgeGraph (if you create a MixedEdgeGraph with only directed and undirected edges it will be recognized as a chain graph in Ananke, and be imported back as such a MixedEdgeGraph), I suppose it's probably fine to leave it in?

adam2392:
I'm not familiar with the MRF part but the rest sounds like a CPDAG. Are they the same thing?

jaron-lee:
Graphically it looks like a CPDAG, yes. If we are just talking the graphical representation it might be fine to subclass the CPDAG object with the added partially directed cycle constraint.

I think we do want these to be distinct classes because e.g. we will have functions that might take a chain graph but not a CPDAG and vice versa, but we would need to tell them apart. Also it would be bad form to suggest to the user via typing that you could e.g. take an output from PC and then pass it to something that parameterizes a chain graph, since this makes no sense.

adam2392:
> I think we do want these to be distinct classes because e.g. we will have functions that might take a chain graph but not a CPDAG and vice versa, but we would need to tell them apart. Also it would be bad form to suggest to the user via typing that you could e.g. take an output from PC and then pass it to something that parameterizes a chain graph, since this makes no sense.

I think the CPDAG implicitly has the acyicity constraint on directed edges. We just don't enforce it because it's expensive to do so. E.g. running a `nx.is_directed_acyclic` function every time an edge is removed/added would be bad.

So I guess another question is what algorithms downstream would a ChainGraph eat but a CPDAG wouldn't and vice versa? Sorry asking all these questions, cuz idk what a chain graph is haha (rip causality community coming up w/ these disparate things)

jaron-lee:
Yeah it's not just acyclicity on the directed edges, it's like A -> B -> C, A - C is forbidden because if we oriented the undirected edge C -> A then you get a cycle. I don't think CPDAGs check for this (and it doesn't matter) but for chain graphs it turns out there is no way to provide a likelihood for such graphs. 

> So I guess another question is what algorithms downstream would a ChainGraph eat but a CPDAG wouldn't and vice versa? Sorry asking all these questions, cuz idk what a chain graph is haha (rip causality community coming up w/ these disparate things)

Yeah one simple example is that a chain graph defines a model (we can give it a likelihood) so you can do things like score it or parameterize it to generate data from it. 

Basically it boils down to the fact that the undirected edge in CPDAG means something very different from in a chain graph so a priori we shouldn't expect that an algorithm should eat both. 

Here's a paper that Ilya was on some time ago https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8313030/

jaron-lee:
Yeah it's not just acyclicity on the directed edges, it's like A -> B -> C, A - C is forbidden because if we oriented the undirected edge C -> A then you get a cycle. I don't think CPDAGs check for this (and it doesn't matter) but for chain graphs it turns out there is no way to provide a likelihood for such graphs. 

> So I guess another question is what algorithms downstream would a ChainGraph eat but a CPDAG wouldn't and vice versa? Sorry asking all these questions, cuz idk what a chain graph is haha (rip causality community coming up w/ these disparate things)

Yeah one simple example is that a chain graph defines a model (we can give it a likelihood) so you can do things like score it or parameterize it to generate data from it. 

Basically it boils down to the fact that the undirected edge in CPDAG means something very different from in a chain graph so a priori we shouldn't expect that an algorithm should eat both. 

Here's a paper that Ilya was on some time ago https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8313030/

adam2392:
I think a CPDAG also implicitly has the no almost directed cycle tho right? E.g. `A -> B -> C` with `A -- C` is not possible based on R2 of Meek's rules. I agree tho we do not "check" this if someone constructs a CPDAG from scratch.

Also, if a check of that sort of acyclicity is required, we do what networkx does and put the responsibility of checking on the user. E.g. in their algorithms section for bipartite graphs, it is on the user to verify that their graph passed in is indeed bipartite.

So I guess a solution is renaming CPDAG to a class that is compatible with both CPDAG and chain graph cuz imo it still sounds quite similar? If we need to add a likelihood, that wouldn't occur on the graph itself I'm guessing, so that should eat this "directed and undirected graph class" and check that it meets the chain graph assumptions by running a check for acylicity and almost acyclicity.

Alternatively, if there is implicitly some structure that we can enforce w/o explicitly running a possibly expensive graph operation every time edges change, then yeah I agree we have two classes: CPDAG and ChainGraph.

jaron-lee:
I think if the philosophy is to put responsibility onto the user then the "directed + bidirected" graph class makes the most sense. I like the solution of having a more generic graph class that can represent both CPDAGs and chain graphs (which is really just a rename of CPDAG to be more inclusive) because that best reflects the underlying situation - the graphical abstraction being useful in different contexts.

I'm not sure how expensive the operation is for checking no partially directed cycles but it's probably at least as expensive as checking ancestors.

I'll create an issue for this separately. In the meantime I will use CPDAG as a stand-in for chain graph.

jaron-lee:
Also the windows and mac CI checks are failing, what was the command that I needed to run to fix that?

adam2392:
> Also the windows and mac CI checks are failing, what was the command that I needed to run to fix that?

Since we want ananke as an "optional" dependency, this can be fixed by moving the import statements for ananke package into the function itself.

adam2392:
> I think if the philosophy is to put responsibility onto the user then the "directed + bidirected" graph class makes the most sense. I like the solution of having a more generic graph class that can represent both CPDAGs and chain graphs (which is really just a rename of CPDAG to be more inclusive) because that best reflects the underlying situation - the graphical abstraction being useful in different contexts.

Yeah from what I loosely understand, a chain graph is a generalization of CPDAG, so it would be nice if we have a good reference explaining all this. Lmk if you have one in mind and I can skim it.

jaron-lee:
There are two kinds of chain graphs that I've seen in the literature. One is the  [Lauritzen, Wermuth, Friedman (LWF) chain graph ](https://www.jstor.org/stable/4616181?casa_token=aaYjqtvwo_UAAAAA%3AwOft-fLbBMixIw3FP2NZrn48DHKctij4hvsmgCIeVpKsvleKEdNSU7AfLXiZFGFeMxQiXIjIEiOvRAuLSX5SM15klcCOnR7g8tGlH5Jg9RoBFNfRLhw)   (see [also](https://www.jstor.org/stable/2245958?casa_token=BBrXFT2cMAEAAAAA%3AtJ_X0aAKB2Zo9sV0KVnd1LQLXhdMtApC9NKkjxh1S54AUNIFf2PB2CMkUUSVHOqc16-s2vq1k4R1cRxymDJjBkPC_anQN4accAAVGXZrPZ2KldUHhCU&seq=5)
while the other is the [Alternative Markov Factorization (AMP) chain graph](https://onlinelibrary.wiley.com/doi/abs/10.1111/1467-9469.00224?casa_token=Gi830aZ4WQwAAAAA:2k2UZd1xu8MN8T5AdA05TM7AOai9YOTxSpP5DmSo30TdMsApQvlphLoqcQPYZDLpOwOFUdzhmvxefA).

These are the original papers that introduce the chain graph as is commonly used in the literature.

There are some more modern papers that focus on the relevance to causal inference - e.g. http://www.di.unito.it/~radicion/tmp_firb__2009/chainGraphs.pdf, which I found easier to understand as an overview of this area.

pywhy_graphs/export/tests/test_ananke.py
```diff
@@ -0,0 +1,123 @@
...
+):
+    # Convert to Ananke graph
+    result = graph_to_ananke(expected_G)
+
+    # Convert back to MixedEdgeGraph
+    graph = ananke_to_graph(result)
+
+    for edge_type, subG in graph.get_graphs().items():
+        assert edge_type in expected_G.edge_types
+        assert nx.is_isomorphic(subG, expected_G.get_graphs(edge_type))
```
adam2392, line 123: Can you use the function inside `pywhy_graphs.testing` to assert the two graphs are isomorphic?

pywhy_graphs/export/ananke.py
```diff
@@ -0,0 +1,125 @@
...
+    graph : pywhy_nx.MixedEdgeGraph
+        The mixed edge causal graph
+    directed_edge_name : str
+        Name of the directed edge, default is directed.
+    bidirected_edge_name : str
+        Name of the bidirected edge, default is bidirected.
+    undirected_edge_name : str
+        Name of the undirected edge, default is undirected.
+
+
```
adam2392, line 28:
```suggestion

```

pywhy_graphs/export/ananke.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    Convert causal graph to Ananke graph. Supports graphs with directed, undirected, and
+    bidirected edges -- including DAGs, ADMGs, and CGs (chain graphs).
+
+    Parameters
+    ----------
+    graph : pywhy_nx.MixedEdgeGraph
+        The mixed edge causal graph
+    directed_edge_name : str
+        Name of the directed edge, default is directed.
```
adam2392, line 22:
```suggestion
        Name of the directed edge, default is 'directed'.
```

pywhy_graphs/export/ananke.py
```diff
@@ -0,0 +1,125 @@
...
+    bidirected_edge_name : str
+        Name of the bidirected edge, default is bidirected.
```
adam2392, line 24:
```suggestion
        Name of the bidirected edge, default is 'bidirected'.
```

pywhy_graphs/export/ananke.py
```diff
@@ -0,0 +1,125 @@
...
+    undirected_edge_name : str
+        Name of the undirected edge, default is undirected.
```
adam2392, line 26:
```suggestion
        Name of the undirected edge, default is 'undirected'.
```

pywhy_graphs/export/ananke.py
```diff
@@ -0,0 +1,125 @@
...
+    Convert Ananke graph to causal graph. Explicitly supports conversion
+    to pywhy_nx.DAG, pywhy_nx.ADMG, and for ananke.graphs.CG to pywhy_nx.CPDAG.
+    Other graph types are converted to a pywhy_nx.MixedEdgeGraph.
+
+    Parameters
+    ----------
+    ananke_graph : Graph
+        The Ananke graph
+    directed_edge_name : str
+        Name of the directed edge, default is directed.
```
adam2392, line 84:
```suggestion
        Name of the directed edge, default is 'directed'.
```

pywhy_graphs/export/ananke.py
```diff
@@ -0,0 +1,125 @@
...
+    bidirected_edge_name : str
+        Name of the bidirected edge, default is bidirected.
```
adam2392, line 86:
```suggestion
        Name of the bidirected edge, default is 'bidirected'.
```

pywhy_graphs/export/ananke.py
```diff
@@ -0,0 +1,125 @@
...
+    undirected_edge_name : str
+        Name of the undirected edge, default is undirected.
```
adam2392, line 88:
```suggestion
        Name of the undirected edge, default is 'undirected'.
```

adam2392 (commented):
Some minor nitpicking, but otw LGTM once these are fixed.

pywhy_graphs/export/tests/test_ananke.py
jaron-lee: Done

adam2392 (approved):

adam2392:
Thanks @jaron-lee !

PR #73, opened 2023-04-07
[ENH] Add ability to check validity of PAGs

### LINKED ISSUE #67: Checking the validity of a constructed PAG
Whenever a PAG is constructed, it is not necessary a valid PAG in the sense of the definition. I.e. it might be a PAG with additional orientations. 

However, for many use-cases, one might want to construct only a valid PAG. So they would like to check its validity in a graphical algorithm.

As suggested by @jdramsey in https://reader.elsevier.com/reader/sd/pii/S0004370208001008?token=1C0B24A74D739F3C8D05A0D2DEC40AA540835B1E58C74D189AE09D9C6FABBF30C941385A4A1F9FC9DFAE1E316E022D1E&originRegion=us-east-1&originCreation=20230320132639, Theorem 2 provides a method for turning a PAG into a MAG. This was originally suggested by Peter Spirtes. 

Imo, this is an extremely valuable algorithm in terms of applications.

Then one can check the validity of the MAG using ancestrality and maximality. If there is a violation of the definition of the MAG, then the PAG is essentially invalid.

This can be implemented as a graph algorithm.

**Additional context**
xref: https://github.com/cmu-phil/tetrad/issues/1556

aryan26roy:
Closes #67

- Adds a function to check the validity of provided PAGs.

## Before submitting

## After submitting

### DIFF
--- a/pywhy_graphs/algorithms/pag.py
+++ b/pywhy_graphs/algorithms/pag.py
@@ -1148,6 +1148,22 @@ def _check_parent_spouse(graph: ADMG):
     return False
 
 
+def _find_all_colliders(graph: ADMG):
+    """Finds and returns a set of all the colliders in a given graph.
+
+    Args:
+        graph (ADMG): The graph.
+    """
+    nodes = graph.nodes
+    print(nodes, "efefeef")
+    out = set()
+    for elem in nodes:
+        parents = list(graph.parents(elem))
+        if len(parents) > 1:
+            out.add(elem)
+    return out
+
+
 def _check_m_seperation(graph: ADMG):
     """Checks to see if every pair of non-adjacent node is m-seperated or not.
 
@@ -1155,14 +1171,16 @@ def _check_m_seperation(graph: ADMG):
         graph (ADMG): The MAG.
     """
     nodes = set(graph.nodes)
-    z = {}
-    # z = _find_all_colliders(graph) #create z containing all the colliders in the graph
+    z = _find_all_colliders(graph)  # create z containing all the colliders in the graph
     for source in nodes:
         nb = set(graph.neighbors(source))
         curset = nodes - nb
         curset.remove(source)  # set of non-adjacent nodes
-
+        if source in z:
+            z.remove(source)
         for dest in curset:
+            if dest in z:
+                z.remove(dest)
             if not m_separated(graph, source, dest, z):
                 return False
     return True
--- a/pywhy_graphs/algorithms/pag.py
+++ b/pywhy_graphs/algorithms/pag.py
@@ -22,6 +22,7 @@ __all__ = [
     "pds_t",
     "pds_t_path",
     "is_definite_noncollider",
+    "is_valid_PAG",
 ]
 
 
@@ -551,6 +552,7 @@ def pds(
     ----------
     .. footbibliography::
     """
+    print("HAHAHAHAHHA")
     if max_path_length is None:
         max_path_length = 1000
 
@@ -910,7 +912,7 @@ def _check_ts_node(node):
         raise ValueError(f"All lag points should be 0, or less. You passed in {node}.")
 
 
-def orient_edges(graph: CPDAG) -> None:
+def _apply_meek_rules(graph: CPDAG) -> None:
     """Orient edges in a skeleton graph to estimate the causal DAG, or CPDAG.
     These are known as the Meek rules :footcite:`Meek1995`. They are deterministic
     in the sense that they are logical characterizations of what edges must be
@@ -922,8 +924,8 @@ def orient_edges(graph: CPDAG) -> None:
     """
     # For all the combination of nodes i and j, apply the following
     # rules.
-    finished = False
-    while not finished:  # type: ignore
+    completed = False
+    while not completed:  # type: ignore
         change_flag = False
         for i in graph.nodes:
             for j in graph.neighbors(i):
@@ -931,15 +933,15 @@ def orient_edges(graph: CPDAG) -> None:
                     continue
                 # Rule 1: Orient i-j into i->j whenever there is an arrow k->i
                 # such that k and j are nonadjacent.
-                r1_add = _apply_meek_rule1(graph, i, j)
+                r1_add = _meek_rule1(graph, i, j)
 
                 # Rule 2: Orient i-j into i->j whenever there is a chain
                 # i->k->j.
-                r2_add = _apply_meek_rule2(graph, i, j)
+                r2_add = _meek_rule2(graph, i, j)
 
                 # Rule 3: Orient i-j into i->j whenever there are two chains
                 # i-k->j and i-l->j such that k and l are nonadjacent.
-                r3_add = _apply_meek_rule3(graph, i, j)
+                r3_add = _meek_rule3(graph, i, j)
 
                 # Rule 4: Orient i-j into i->j whenever there are two chains
                 # i-k->l and k->l->j such that k and j are nonadjacent.
@@ -949,11 +951,11 @@ def orient_edges(graph: CPDAG) -> None:
                 if any([r1_add, r2_add, r3_add]) and not change_flag:
                     change_flag = True
         if not change_flag:
-            finished = True
+            completed = True
             break
 
 
-def _apply_meek_rule1(graph: CPDAG, i: str, j: str) -> bool:
+def _meek_rule1(graph: CPDAG, i: str, j: str) -> bool:
     """Apply rule 1 of Meek's rules.
     Looks for i - j such that k -> i, such that (k,i,j)
     is an unshielded triple. Then can orient i - j as i -> j.
@@ -980,7 +982,7 @@ def _apply_meek_rule1(graph: CPDAG, i: str, j: str) -> bool:
     return added_arrows
 
 
-def _apply_meek_rule2(graph: CPDAG, i: str, j: str) -> bool:
+def _meek_rule2(graph: CPDAG, i: str, j: str) -> bool:
     """Apply rule 2 of Meek's rules.
     Check for i - j, and then looks for i -> k -> j
     triple, to orient i - j as i -> j.
@@ -990,18 +992,18 @@ def _apply_meek_rule2(graph: CPDAG, i: str, j: str) -> bool:
     # Check if i-j.
     if graph.has_edge(i, j, graph.undirected_edge_name):
         # Find nodes k where k is i->k
-        succs_i = set()
+        child_i = set()
         for k in graph.successors(i):
             if not graph.has_edge(k, i, graph.directed_edge_name):
-                succs_i.add(k)
+                child_i.add(k)
         # Find nodes j where j is k->j.
-        preds_j = set()
+        parent_j = set()
         for k in graph.predecessors(j):
             if not graph.has_edge(j, k, graph.directed_edge_name):
-                preds_j.add(k)
+                parent_j.add(k)
 
         # Check if there is any node k where i->k->j.
-        candidate_k = succs_i.intersection(preds_j)
+        candidate_k = child_i.intersection(parent_j)
         # if the graph has excluded triples, we would check at this point
         if graph.excluded_triples:
             # check if the triple is in the graph's excluded triples
@@ -1018,7 +1020,7 @@ def _apply_meek_rule2(graph: CPDAG, i: str, j: str) -> bool:
     return added_arrows
 
 
-def _apply_meek_rule3(graph: CPDAG, i: str, j: str) -> bool:
+def _meek_rule3(graph: CPDAG, i: str, j: str) -> bool:
     """Apply rule 3 of Meek's rules.
     Check for i - j, and then looks for k -> j <- l
     collider, and i - k and i - l, then orient i -> j.
@@ -1114,7 +1116,7 @@ def is_valid_PAG(graph: PAG) -> bool:
             for (u, v) in undedges:
                 temp_cpdag.remove_edge(u, v, temp_cpdag.undirected_edge_name)
                 temp_cpdag.add_edge(u, v, temp_cpdag.directed_edge_name)
-                orient_edges(temp_cpdag)
+                _apply_meek_rules(temp_cpdag)
                 print(temp_cpdag.directed_edges)
                 break
         else:
--- a/pywhy_graphs/algorithms/pag.py
+++ b/pywhy_graphs/algorithms/pag.py
@@ -1,6 +1,6 @@
 import logging
 from collections import deque
-from itertools import chain
+from itertools import chain, combinations
 from typing import List, Optional, Set, Tuple
 
 import networkx as nx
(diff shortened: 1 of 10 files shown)


aryan26roy:
@adam2392 I made this draft PR since I am uncertain about a couple of things:

- Is this the right place to put this function? (I thought of making it a method of the PAG class but you said you intend to do away with these specific classes, so chose to put it in this file instead.)
- Right now I am only checking if the provided PAG has circle edges, is this enough?
- I implemented the first step of the theorem 2 from the linked paper. The second step requires getting rid of all unshielded colliders, for this I have to add edges between several node. What should these edges be? Directed? Undirected? Bidirected? If they are directed, in what direction?

aryan26roy:
I found the tetrad [code](https://github.com/cmu-phil/tetrad/blob/40799a90da8d878c3f67305b84d78dde94934fb9/tetrad-lib/src/main/java/edu/cmu/tetrad/search/SearchGraphUtils.java#L701) for turning a PAG into an MAG!




> No, you have to follow the theorem linked and then the definition of the MAG. Just cuz you have or don't have circle edges does not imply anything.

I meant to ask if this check is enough as a pre-cursor to the actual validity check. I am assuming that a graph cannot be a PAG if it doesnt have any circle edges, and if so, it is automatically classified as an invalid PAG.



> This step may require some generic graph traversal... this issue might actually be harder than I thought, but it's just a standard graph BFS/DFS traversal if you're familiar with graph algorithms. E.g. see the other algorithms, and what they do.

I am! (I took multiple classes on graphs and I have done a past project where I implemented some of the graph traversal algorithms for robotics application.) I still expect to encounter some difficulties though. I will look through the tetrad code and let you know if I am unable to understand what to do exactly after that.

aryan26roy:
Ok so I looked it up and it turns out that a graph can be a PAG even if it doesn't have any circle edges.

pywhy_graphs/algorithms/pag.py
```diff
@@ -934,15 +934,23 @@ def is_valid_PAG(graph: PAG) -> bool:
 
     to_remove = []
     to_reorient = []
+    to_replace = []
     for u, v in cedges:
         if (v, u) in dedges:
             to_remove.append((u, v))
         elif (v, u) not in cedges:
             to_reorient.append((u, v))
+        elif (v, u) in cedges and (v, u) not in to_replace:
```
adam2392, line 943: The `not in to_replace` does not scale well. You should use a different data structure to check inclusion (e.g. set, or dict)

adam2392: I would add some in-line comment to describe this sequence of if/else statements.

adam2392:
Feel free to ping me when/if there's any issues or you need a review.

pywhy_graphs/algorithms/pag.py
aryan26roy: Makes sense.

aryan26roy: Sure.

aryan26roy:
@adam2392 sorry for the delay but I have been stuck with a work emergency. Will get back to this issue from Wednesday.


> Feel free to ping me when/if there's any issues or you need a review.

Occasionally I have a problem understanding some theoretical things and I feel like GitHub is not the place clarify them. Is there a better way to communicate with you (like email or slack?) or would like me to ask all of those question here as well?

adam2392:
> @adam2392 sorry for the delay but I have been stuck with a work emergency. Will get back to this issue from Wednesday.

Sounds good no rush!

> Occasionally I have a problem understanding some theoretical things and I feel like GitHub is not the place clarify them. Is there a better way to communicate with you (like email or slack?) or would like me to ask all of those question here as well?

Feel free to summarize them here, in case other devs following e.g. @jaron-lee. If it warrants a call, then we can hop on discord. Are you on the discord for pywhy?

aryan26roy:
> Feel free to summarize them here, in case other devs following e.g. @jaron-lee. If it warrants a call, then we can hop on discord. Are you on the discord for pywhy?

I am not. But I will join now. Did not know a discord server existed for pywhy (I was expecting a slack channel but I like Discord more anyway :-))

robertness (commented):
I suggest adding sub-functions to `is_valid_PAG` to make it more modular.

aryan26roy:
@robertness you're right. I have been planning to make major chunks of the implementation modular (starting from the meek rules).

pywhy_graphs/algorithms/pag.py
```diff
@@ -908,3 +908,232 @@ def _check_ts_node(node):
         )
     if node[1] > 0:
         raise ValueError(f"All lag points should be 0, or less. You passed in {node}.")
+
+
+def orient_edges(graph: CPDAG) -> None:
```
adam2392, line 913: Probably don't just copy/paste the code directly. Prolly should be renamed, etc.

E.g. `orient_edges` is sort of a bad name, cuz it is in the `pag.py` file and it works on a CPDAG. Perhaps, you rename it to a private function called `_apply_meek_rules`

aryan26roy: Sure.

aryan26roy:
@adam2392 I am trying to import the function and use it in a script like this:
```pywhy_graphs.algorithms.is_valid_PAG(pag)```
The output is this error:
>AttributeError: module 'pywhy_graphs.algorithms' has no attribute 'is_valid_PAG'

Do you know why?

**Comment by adam2392** [comment]
> @adam2392 I am trying to import the function and use it in a script like this: `pywhy_graphs.algorithms.is_valid_PAG(pag)` The output is this error:
> 
> > AttributeError: module 'pywhy_graphs.algorithms' has no attribute 'is_valid_PAG'
> 
> Do you know why?

You have to expose it to the import module by adding an entry to `__all__` inside the `algorithms/pag.py` file at the top.

File: pywhy_graphs/algorithms/pag.py [hunk 3]
```diff
@@ -908,3 +910,277 @@ def _check_ts_node(node):
         )
     if node[1] > 0:
         raise ValueError(f"All lag points should be 0, or less. You passed in {node}.")
+
+
+def _apply_meek_rules(graph: CPDAG) -> None:
+    """Orient edges in a skeleton graph to estimate the causal DAG, or CPDAG.
+    These are known as the Meek rules :footcite:`Meek1995`. They are deterministic
```
adam2392, line 917:
```suggestion
    """Orient edges in a skeleton graph to estimate the causal DAG, or CPDAG.

    These are known as the Meek rules :footcite:`Meek1995`. They are deterministic
```

pywhy_graphs/algorithms/pag.py
```diff
@@ -908,3 +910,277 @@ def _check_ts_node(node):
...
+    in the sense that they are logical characterizations of what edges must be
+    present given the rest of the local graph structure.
+    Parameters
```
adam2392, line 920:
```suggestion
    present given the rest of the local graph structure.

    Parameters
```

pywhy_graphs/algorithms/pag.py
```diff
@@ -551,6 +552,7 @@ def pds(
     ----------
     .. footbibliography::
     """
+    print("HAHAHAHAHHA")
```
jaron-lee, line 555: I assume this is just a debugging tool but should be removed before committing

aryan26roy: Sure!

aryan26roy:
@adam2392 I think I have constructed the MAG correctly (I think the rule 4 is correctly implemented). Now how do I check the validity of an MAG? Can you explain it to me in a detailed way?

adam2392:
> @adam2392 I think I have constructed the MAG correctly (I think the rule 4 is correctly implemented). Now how do I check the validity of an MAG? Can you explain it to me in a detailed way?

What regarding the MAG validity do you have a question on from the paper's definition?

Some things you might want to consider are to start adding unit tests for this function, so that way we can explicitly review when the function returns True/False. You can see how in the other functions, we explicitly construct graphs that meet some condition, and then does not.

adam2392:
Reposting here question from @aryan26roy:

> I am going through the definition of MAGs in the paper and I have some questions about it.
-  What is an almost directed cycle?
- Does "spouse" mean two nodes that have a bidirected edge?
- I am a bit unsure about the m-seperation. Do you have a resource that has implemented a check for m-separation?  Or some other easier way understanding it?

In the paper:

- "An almost directed cycle occurs when B ↔ A is in G and A ∈ AnG (B)"
- "A is called a spouse of B (and B a spouse of A) if A ↔ B is in G"

Re m-separation: https://www.pywhy.org/pywhy-graphs/dev/generated/pywhy_graphs.networkx.m_separated.html#pywhy_graphs.networkx.m_separated

aryan26roy:
Ok so I seem to have a picture of what I need to implement:
- A function call (assuming it has already been implemented) that checks for directed cycles.
- A function for checking of almost directed cycles (i.e directed cycles with possibly one edge being bidirected).
- A function for checking for parents and spouses of nodes with undirected edges between them.
- A function call for asserting m-separation between any two set of nodes.

aryan26roy:
@adam2392 The function that you gave a link to, requires 'z' a set of nodes that m-separate the two nodes. Do I pass an empty set there?

adam2392:
> @adam2392 The function that you gave a link to, requires 'z' a set of nodes that m-separate the two nodes. Do I pass an empty set there?

What are you trying to do?

aryan26roy:
Oh I seem to have confused the purpose of the function. Essentially I will have to implement a function that given a path 'p' between each pair of nodes X and Y finds a set Z where:
-  X, Y /∈ Z
- every non-collider on p is not a member of Z;
-  every collider on p has a descendant in Z.
If Z turns out to be empty, then X and Y are not m-separable? Did I get it right?

aryan26roy:
@adam2392 What I am trying to ask is, given that the paper says:
```An ancestral graph is said to be maximal if for any two non-adjacent vertices, there is a set of vertices that m-separates them.```
The set of vertices is 'z' for that function. How do I get 'z'? And since the paper (and the function) is talking in terms of sets of nodes X and Y, how do I obtain this split? (One idea is that X contains one node and Y contains all the non-adjacent nodes from that one node)

**Comment by adam2392** [comment]
> `An ancestral graph is said to be maximal if for any two non-adjacent vertices, there is a set of vertices that m-separates them.`
> The set of vertices is 'z' for that function. How do I get 'z'? And since the paper (and the function) is talking in terms of sets of nodes X and Y, how do I obtain this split? (One idea is that X contains one node and Y contains all the non-adjacent nodes from that one node)

Naively check all possible combinations of subsets of vertices. There's probably a more efficient way of doing this I imagine tho.

**Comment by adam2392** [comment]
I think the easier thing might be to start with a few tests where this should work and where it shouldn't work. You can start w/ small PAGs and use tetrad GUI to construct examples that you then add into a unit-test.

**Comment by aryan26roy** [comment]
Ah! Thank you for the suggestion @adam2392. I will have to setup tetrad before that but this will make my job much easier.

**Comment by aryan26roy** [comment]
@adam2392 I am facing a weird issue. In the middle of the function, I construct a temporary CPDAG. When I print out the edges of the CPDAG, this is what I get:
```
{'directed': OutEdgeView([('w', 'z'), ('z', 'x'), ('x', 'xy'), ('xy', 'w')]), 'undirected': EdgeView([])}
```
But when I draw the graph, I get two different graphs on different runs (both are wrong too):

weird2

weird

Note that the edge list remains the same in both the graphs.

adam2392:
> @adam2392 I am facing a weird issue. In the middle of the function, I construct a temporary CPDAG. When I print out the edges of the CPDAG, this is what I get:
> 
> ```
> {'directed': OutEdgeView([('w', 'z'), ('z', 'x'), ('x', 'xy'), ('xy', 'w')]), 'undirected': EdgeView([])}
> ```
> 
> But when I draw the graph, I get two different graphs on different runs (both are wrong too):
> 
> weird2
> 
> weird
> 
> Note that the edge list remains the same in both the graphs.

Do you have some reproducing MVP code for this? This seems like a bug in the `draw` function then

aryan26roy:
@adam2392 I am sharing the MVP code for this through [pastebin](https://pastebin.com/ELqNjY0M) as it is too long for a github comment.
I am planning to attend the last one but something else came in the way. I will be there for the next one!

adam2392:
> @adam2392 I am sharing the MVP code for this through [pastebin](https://pastebin.com/ELqNjY0M) as it is too long for a github comment.
> I am planning to attend the last one but something else came in the way. I will be there for the next one!

For future, it might be helpful to keep the code a bit shorter to make the error apparent.

I took a look tho and it seems like there's probably just a bug in the CPDAG drawing. It is drawing extra edges it should not be.

aryan26roy:
@adam2392 I have three questions for you.
- The paper says for m-seperation you check if every collider in the path has a descendant in the set 'z' or not. While the function you linked checks if the collider itself is in 'z' or not. Which is the correct check?
- To check for m-seperation what I did was instead of making random sets to use as 'z' I made a set of all the colliders in the graph. Now all the colliders in any path are necessarily in this set and all the non-colliders are necessarily not. I think this is a better way doing it, what do you think?
- You said that I will have to reconstruct the PAG from the MAG. Is that really important? Because my code does not track the changes made to the graph during the application of the meek rules. Isn't checking the validity of the MAG enough?

aryan26roy:
@adam2392 As for your question, yes, I am able to replicate the bug with just two three nodes.

adam2392:
> @adam2392 I have three questions for you.
> 
> * The paper says for m-seperation you check if every collider in the path has a descendant in the set 'z' or not. While the function you linked checks if the collider itself is in 'z' or not. Which is the correct check?

The function implements it correctly and checks for more than just collider status. https://github.com/py-why/pywhy-graphs/blob/8658c5b86a221285616f7bd944bd262614c947a7/pywhy_graphs/networkx/algorithms/causal/m_separation.py#L151-L165

Try looking up d-separation materials and reading up on why descendants open up collider paths.

> * To check for m-seperation what I did was instead of making random sets to use as 'z' I made a set of all the colliders in the graph. Now all the colliders in any path are necessarily in this set and all the non-colliders are necessarily not. I think this is a better way doing it, what do you think?

Why not just use `minimal_m_separator`, which performs searches to determine if a minimal m-separator exists?

@jaron-lee does `minimal_m_separator` work to just check if two nodes have an m-separating set? It just does the extra work to make sure it is also minimal right? But if two nodes are m-separated, then `minimal_m_separator` should return some set always?

> * You said that I will have to reconstruct the PAG from the MAG. Is that really important? Because my code does not track the changes made to the graph during the application of the meek rules. Isn't checking the validity of the MAG enough?

This is not true because the PAG reconstructed from the MAG may not match the PAG you fed in.

I think if you implement a few positive/negative test cases starting w/ what is easy, this will be easier to discuss. Do you have such unit tests yet? Do you mind pushing them up to the PR?

adam2392:
> @adam2392 As for your question, yes, I am able to replicate the bug with just two three nodes.

Can you open a GH issue related to this? and paste the MVP code sample?

aryan26roy:
> Why not just use `minimal_m_separator`, which performs searches to determine if a minimal m-separator exists?

@adam2392 I didn't know of this function. I will use it now.



> This is not true because the PAG reconstructed from the MAG may not match the PAG you fed in.

How come? I was under the belief that when you say 'reconstruct' the PAG, you want me to re-trace the steps followed throughout the function backwards. If I do that correctly, won't the reconstructed PAG always match the input PAG? And if not this, then what do you mean by reconstruct?

> I think if you implement a few positive/negative test cases starting w/ what is easy, this will be easier to discuss. Do you have such unit tests yet? Do you mind pushing them up to the PR?

I do not have such test cases yet. Do you know where I can find test cases that have pre-tagged valid and non-valid PAGs?

adam2392:
> I do not have such test cases yet. Do you know where I can find test cases that have pre-tagged valid and non-valid PAGs?

If you use Tetrad and draw an arbitrary graph with circle endpoints and then run the check via the GUI (I think it's called check valid graph coloring or something like that), you will be able to construct various cases.

pywhy_graphs/algorithms/tests/test_pag.py
adam2392, line 1: Great start on the unit-tests. 

You can further think of unit-tests by taking a look at each of the FCI rules and draw in tetrad a graph that is sort of opposite those rules, and then you can add a separate unit-test checking each one.

E.g.

X <-o Y <-o Z <-o A

is invalid and so is

X <- Y <- Z <-o A; Z <-o B

because Z <-o A should be Z<-A due to R1 of FCI. And so on.

adam2392: Note: this might not cover everything in the algorithm, but then this is a great step in the right direction

aryan26roy: @adam2392 It took me a while to figure out Tetrad.

How many different tests do you want me to add?

adam2392: And it will help us narrow down the possible runtime error-surfaces. For graph stuff, it's hard to test every single edge case, but good to test things that we know are definitely wrong/correct.

aryan26roy: @adam2392 is this enough?

aryan26roy:
@adam2392 this branch has become too convoluted for me to keep everything straight. Do you mind if I delete this and start fresh on a new branch?

adam2392:
Yes that is a perfectly fine and acceptable practice

PR #78, opened 2023-05-22
[ENH] Add ability to determine whether an inducing path exists between two nodes

### LINKED ISSUE #70: A function to determine whether an inducing path exists between two nodes
An inducing path defined in [1]_ is central to extending reasoning of DAGs to MAG/PAGs, so it makes sense that we should expose an algorithm for computing inducing paths

Implement a function in the `algorithms` submodule with the API:

`inducing_path(G, node_x, node_y, relative_to=None) -> Tuple[bool, path]`, which checks whether a path exists and if it does, it returns the path. We do not necessarily need to check all paths though(?) That is up for discussion I guess.

**Additional context**
[1]: https://reader.elsevier.com/reader/sd/pii/S0004370208001008?token=4B17FC536FD415013F477F7092BFDCC8861C4A6C05FB9562DA2A15DE777CD23D132FA76C0ECAE9B59FE84EADAC54959F&originRegion=us-east-1&originCreation=20230322132043

This will then enable someone to implement the `DAGtoMAG` algorithm.

aryan26roy:
Closes #70 

- Adds a function to determine whether an inducing path exists between two nodes

## Before submitting

## After submitting

### DIFF
--- a/pywhy_graphs/algorithms/pag.py
+++ b/pywhy_graphs/algorithms/pag.py
@@ -908,3 +908,32 @@ def _check_ts_node(node):
         )
     if node[1] > 0:
         raise ValueError(f"All lag points should be 0, or less. You passed in {node}.")
+
+
+def inducing_path(G, node_x, node_y, relative_to=None):
+    """Checks if an inducing path exists between node_x and node_y and if it does returns it.
+
+    Args:
+        G : Graph
+            The graph.
+        node_x : node
+            The source node.
+        node_y : node
+            The final node.
+        relative_to : set
+            The set relative to which the path has to be found. Defaults to None.
+
+
+    Returns:
+        path : Tuple[bool, path]
+            A tuple containing a bool and a path if the bool is true.
+    """
+
+    nodes = G.nodes
+    if node_x not in nodes or node_y not in nodes:
+        raise ValueError("The provided nodes are not in the graph.")
+
+    path = []  # this will contain the path.
+    path.append(node_x)
+
+    return (False, [])


aryan26roy:
The tetrad [implementation](https://github.com/cmu-phil/tetrad/blob/2fee3e19bfd79964d872a87845da52714880b8a1/tetrad-lib/src/main/java/edu/cmu/tetrad/graph/Paths.java#L722) for reference.

aryan26roy:
@adam2392 one more question:
- Neither the children method nor the parents method finds nodes connected through a bi-directed edge. Is there a function that finds child and parent nodes connected through bi-directed edges as well?

adam2392:
> @adam2392 one more question:
> 
> * Neither the children method nor the parents method finds nodes connected through a bi-directed edge. Is there a function that finds child and parent nodes connected through bi-directed edges as well?

You can query the networkx subgraph. Bidirected edges are stored using a `nx.Graph`.

aryan26roy:
@adam2392 so it turns out the "parents" method doesn't work on the bidirected sub-graph. I am using the neighbors method.(since in a bidirected graph a neighbor is also a parent.)

adam2392:
can you add test cases as well?

You can just construct various graphs on paper that follow the definition or not to test.

aryan26roy:
@adam2392 is this an appropriate amount of tests?

adam2392 (commented):
Nice! I would say enough tests to convince the reviewer your implementation is correct ;)

I would split this into two tests. 

One where L and S are possibly defined. And another where they are the empty set. 

You can then have an inducing path if 

X-> Y and X<->Y

aryan26roy:
@adam2392 The implementation seems to be correct and all the tests are passing as well. Want to review?

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,192 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+            The destination node.
+        L : set
+            The set containing every non-collider on the path.
+        S:  set
+            The set containing every collider on the path that is not an ancestor of the endpoints.
+
+
+    Returns:
+        path : Tuple[bool, path]
+            A tuple containing a bool and a path if the bool is true.
```
adam2392, line 481: Technically a path is always returned right? It just will be empty?

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,192 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+                G, node_x, node_y, L, S, visited, xyancestors, elem
+            )
+            if path_exists:
+                path.append(cur_node)
+                path.extend(temp_path)
+                break
+    return (path_exists, path)
+
+
+def inducing_path(G, node_x, node_y, L=set(), S=set()):
```
adam2392, line 463: You should not initialize to a set(), or any mutable object. https://stackoverflow.com/questions/1534407/python-object-initialization-bug-or-am-i-misunderstanding-how-objects-work

A pattern is initializing to `None` and then using if/else conditionals to set the default parameters. You should note this in the docstring also.

pywhy_graphs/algorithms/tests/test_generic.py
```diff
@@ -39,3 +40,73 @@ def test_convert_to_latent_confounder(graph_func):
...
+
+    S = {"Y", "Z"}
+    L = set()
+
+    assert not pywhy_graphs.inducing_path(admg, "X", "K", L, S)[
+        0
+    ]  # A non-collider on the path is not in S
+
+
+def test_inducing_path_wihtout_LandS():
```
adam2392, line 95: I would also add the test case:

```
A -> B -> C -> D,
B <-> C,
D <-> C
```

and then variants of that where one of the nodes does not satisfy the inducing path definition.

pywhy_graphs/algorithms/tests/test_generic.py
```diff
@@ -39,3 +40,73 @@ def test_convert_to_latent_confounder(graph_func):
     G.remove_edge(3, 1, G.bidirected_edge_name)
     G.add_edge(3, 1, G.directed_edge_name)
     assert pywhy_graphs.is_node_common_cause(G, 3)
+
+
+def test_inducing_path():
```
adam2392, line 45: Same comment as below. I would also add varying cases of graphs also with multiple edges (this is possible in a causal diagram, but not in an ancestral graph like MAG/PAG)

E.g.
```
A -> B -> C -> D,
B <-> C,
```

but with L = {'C'}, will make it an inducing path. But with L = {}, will not be an inducing path.

adam2392 (changes requested):
This is great progress! I added some high-level comments to the tests and also the function API.

Re docstrings: please use numpydoc style

I will review more in-depth once those tests pass because rn upon looking at the tests, I am not 100% convinced it will work as expected yet.

pywhy_graphs/algorithms/tests/test_generic.py
aryan26roy: @adam2392 isn't L supposed to contain non-colliders? If so, to make it a inducing path,  L = {'B', 'C'}, right?

adam2392: Yes but C is not a collider going from A to D. Going from A to D, each node in my example is a collider, descendant of a member of S, or a member of L.

B == collider
C == part of L
A == endpoint
D == endpoint

adam2392: You can confirm this by trying it out in tetrad too

adam2392: > @adam2392 isn't L supposed to contain non-colliders? If so, to make it an inducing path, L = {'B', 'C'}, right?

This also makes it an inducing path. Intuitively, A and D are d-connected no matter what. If you say L = {'B', 'C'}, then B and C are not able to be conditioned on. A and D are d-connected then.

In my example, if L = {'C'}, same thing. C is unable to be conditioned on, so A and D are d-connected because if you don't condition on B, the path is open. If you condition on B, the path is still open.

aryan26roy:
@adam2392 I think I added all the tests you asked for. They worked without changing anything in the code. Can you take a look?

pywhy_graphs/algorithms/tests/test_generic.py
aryan26roy: Ah thanks for the explanation but I had asked that question out of confusion. I didn't realize B was a collider.

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,207 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    L : set
+        The set containing every non-collider on the path. Defaults to an empty set.
+    S:  set
+        Contains all colliders on the path. Defaults to an empty set.
+
+
+    Returns
+    -------
+    path : Tuple[bool, path]
+        A tuple containing a bool and a path if the bool is true.
```
adam2392, line 490: What if bool is False? What is returned?

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,207 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+        The graph.
+    node_x : node
+        The source node.
+    node_y : node
+        The destination node.
+    L : set
+        The set containing every non-collider on the path. Defaults to an empty set.
+    S:  set
+        Contains all colliders on the path. Defaults to an empty set.
+
```
adam2392, line 485:
```suggestion
```

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,207 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    Parameters
+    ----------
+    G : Graph
+        The graph.
+    node_x : node
+        The source node.
+    node_y : node
+        The destination node.
+    L : set
+        The set containing every non-collider on the path. Defaults to an empty set.
```
adam2392, line 482:
```suggestion
        The ``L`` set that contains nodes that are ignored on the path. Defaults to an empty set. See Notes for details.
```

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,207 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    S:  set
+        Contains all colliders on the path. Defaults to an empty set.
```
adam2392, line 484:
```suggestion
        The ``S`` set on the path containing nodes that are always conditioned on. Defaults to an empty set. See Notes for details.
```

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,207 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+            if path_exists:
+                path.append(cur_node)
+                path.extend(temp_path)
+                break
+    return (path_exists, path)
+
+
+def inducing_path(G, node_x, node_y, L=None, S=None):
+    """Checks if an inducing path exists between node_x and node_y and if it does returns it.
+
```
adam2392, line 472: Add a reference to the Zhang paper on the completeness of FCI, where inducing path is defined. We use sphinx-bibtex, so some other docstrings show the pattern.

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,207 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+        The set containing every non-collider on the path. Defaults to an empty set.
+    S:  set
+        Contains all colliders on the path. Defaults to an empty set.
+
+
+    Returns
+    -------
+    path : Tuple[bool, path]
+        A tuple containing a bool and a path if the bool is true.
+    """
```
adam2392, line 491: Add a `Notes` section describing i) what is the intuition of an inducing path and ii) describing how the hyper parameters ``L`` and ``S`` are used.

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,207 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    Returns
+    -------
+    iscollider : bool
+        Bool is set true if the node is a collider, false otherwise.
+    """
+    parents = _find_directed_parents(G, node)
+
+    if len(parents) > 1:
+        return True
+    return False
```
adam2392, line 407: The way a collider is defined is given (X, Y, Z), Y is a collider in G if X and Y are adjacent, and Y and Z are adjacent and X *-> Y and Y <-* Z.

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,207 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
                     nextlevel[w] = 1
         level += 1
     return paths
+
+
+def _find_directed_parents(G, node):
+    """Finds the parents of a node in the directed and the bidirected subgraphs.
```
adam2392, line 340: This is a weird name for the function. When I review it, it sounds like it should only find the parents in the directed subgraph of `node`, but in fact it finds the parents in the directed subgraph + the adjacent neighbors.

To make things more maintainable, it is better to have a separate function for:

1. parents in the directed subgraph
2. neighbors in the bidirected subgraph

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,207 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    node : node label
+        The node for which we have to find the parents.
+
+    Returns
+    -------
+    out : set
+        The parents of the provided node.
+    """
+
+    if not isinstance(G, CPDAG):
```
adam2392, line 355: Not sure why this check is needed.

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,207 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    visited.add(node_x)
+
+    path = []  # this will contain the path.
+
+    xanc = _find_directed_parents(G, node_x)
+    yanc = _find_directed_parents(G, node_y)
+
+    xyancestors = xanc.union(yanc)
+
+    children = _find_directed_children(G, node_x)
```
adam2392, line 517: This is not correct:

```
    admg = ADMG()

    admg.add_edge("Y", "X", admg.directed_edge_name)
    admg.add_edge("Z", "Y", admg.bidirected_edge_name)
    admg.add_edge("Z", "H", admg.bidirected_edge_name)

    # this is an inducing path if L ={'Y'} and S = {}, yet it fails
    # X <- Y <-> z <-> H
    S = set()
    L = {'Y'}
    assert pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]
```

adam2392 (changes requested):
Left some comments. I think this is not matching the Java implementation, or the definition as I point out with a counter-example. I think making the abstractions simpler and using proper names may make the whole thing a bit easier to implement in a modular fashion.

Here is a breakdown of pseudo-code for the inducing path algorithm:

Consider starting node X and ending node Y. Add all adjacent neighbors of node X to the stack.

If we encounter a node that we have not seen yet, then we ignore it if:
1. it is not on a path to Y not through X (possibly a naive thing to do would just be check the converted undirected graph for a path): If it's not even connected to Y, except through X, then this is not a path, let alone an inducing path.

Now the new node is forsure at least on a path (directed, or bidirected, or both) to Y. The new node is a candidate for adding to the path if:

1. if node is in L
2. if node is a collider, check that it is an ancestor of X, Y, or some member of S.

If any of the above two conditions, add the node to the path and also add the node to the stack, else, it is not a candidate for the path, and we move on. Continue until Y is found and return the path, or until stack is empty. Note: I don't think using a stack is necessarily the right move, but this is just for illustration purposes.

Looking at the Java implementation, this is at a high-level pretty similar. https://github.com/cmu-phil/tetrad/blob/2fee3e19bfd79964d872a87845da52714880b8a1/tetrad-lib/src/main/java/edu/cmu/tetrad/graph/Paths.java#L722

Comparing the tetrad and Python implementation, there are a few key differences already that seem to make all this harder:
1. collider check is not defined in your implementation correctly
2. checking ancestors can be done on the directed subgraph using: https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.ancestors.html

Probably more, but hopefully you get the idea.

pywhy_graphs/algorithms/generic.py
aryan26roy: I am not sure what is wrong with this implementation. Is it the case that it can only be called a collider if and only if there are 2 nodes that have a directed edge pointing towards the third node?

pywhy_graphs/algorithms/generic.py
aryan26roy: Because a CPDAG does not have a bidirected subgraph. At least not in the class defintion.

pywhy_graphs/algorithms/generic.py
aryan26roy: Wait, in the graph that you constructed:


> # X <- Y <-> z <-> H

The directed edge is Y -> X . Is this a mistype or are you expecting a directed edge to mean a connection, no matter what way it is oriented? I thought an edge was two way traversable only if it was a bi-directed edge.

pywhy_graphs/algorithms/generic.py
adam2392: Yep! A collider is a collider if there is two arrow endpoints. Usually it is defined as a function of three nodes. 

So 

Is_collider(G, a,b,c)

Checks if say b is a collider by determining if both a and c have an arrow pointing to b. This can be via a directed edge or bidirected edge.

pywhy_graphs/algorithms/generic.py
adam2392: Let's assume the graph is a mixed edge graph. If so, it may or may not have any of these edge subgraphs.

pywhy_graphs/algorithms/generic.py
adam2392: Sorry this is a typo. I'll clarify on my computer later.

adam2392:
```
# X <- Y <-> Z <-> H; Z -> X
    admg = ADMG()
    admg.add_edge("Y", "X", admg.directed_edge_name)
    admg.add_edge("Z", "X", admg.directed_edge_name)
    admg.add_edge("Z", "Y", admg.bidirected_edge_name)
    admg.add_edge("Z", "H", admg.bidirected_edge_name)

    # not an inducing path, since Y is not a collider and Y is not part of L
    S = {}
    L = {}
    assert not pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]

    # now an inducing path, since Y is not a collider, but is part of L
    L = {'Y'}
    assert pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]

    # X <-> Y <-> Z <-> H; Z -> X
    admg = ADMG()
    admg.add_edge("Y", "X", admg.bidirected_edge_name)
    admg.add_edge("Z", "X", admg.directed_edge_name)
    admg.add_edge("Z", "Y", admg.bidirected_edge_name)
    admg.add_edge("Z", "H", admg.bidirected_edge_name)

    # not an inducing path, since Y is not an ancestor of X, H, or S
    S = {}
    L = {}
    assert not pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]

    # still not an inducing path, since Y is a collider
    L = {'Y'}
    assert not pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]

    # now add an edge Y -> A
    admg.add_edge("Y", "A", admg.directed_edge_name)

    # an inducing path, since Y is a collider and is an ancestor of X, H, or S
    L = {}
    S = {'A'}
    assert pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]

    # an inducing path, since Y is a collider and is an ancestor of X, H, or S
    L = {}
    S = {'Y'}
    assert pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]
```

These tests act as a lead up to various examples that should pass but does not based on the current implementation.

adam2392:
@aryan26roy thanks for iterating on this! Feel free to come to the Discord OH, or other time periods and ping the channel so we know we should discuss this if you have any questions.

Inducing paths are somewhat abstract at first tbh, so once the definition is clear, I think this will come out cleanly.

pywhy_graphs/algorithms/generic.py
aryan26roy: >  now an inducing path, since Y is not a collider, but is part of L
>     L = {'Y'}
>     assert pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]

@adam2392 There is still no directed path from X to H. Should this example still pass?

aryan26roy: >     # X <-> Y <-> Z <-> H; Z -> X
>     admg = ADMG()
>     admg.add_edge("Y", "X", admg.bidirected_edge_name)
>     admg.add_edge("Z", "X", admg.directed_edge_name)
>     admg.add_edge("Z", "Y", admg.bidirected_edge_name)
>     admg.add_edge("Z", "H", admg.bidirected_edge_name)
> 
>     # not an inducing path, since Y is not an ancestor of X, H, or S
>     S = {}
>     L = {}
>     assert not pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]

Doesn't the bi-directed edge (X <-> Y) mean that Y is an ancestor of X? I have been taking it to mean that.

adam2392: > Doesn't the bi-directed edge (X <-> Y) mean that Y is an ancestor of X? I have been taking it to mean that.

No ancestor is only thru directed edges.

adam2392: > > now an inducing path, since Y is not a collider, but is part of L
> > L = {'Y'}
> > assert pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]
> 
> @adam2392 There is still no directed path from X to H. Should this example still pass?

According to the definition, yes:

```
# X <- Y <-> Z <-> H; Z -> X
```

Y is a non-collider and part of L. Z is a collider and an ancestor of X, Y, or S. X and H are endpoints. Lmk if that doesn't make sense considering the definition of inducing path.

aryan26roy: >     #  X <- Y <-> Z <-> H; Z -> X
>     admg = ADMG()
>     admg.add_edge("Y", "X", admg.directed_edge_name)
>     admg.add_edge("Z", "X", admg.directed_edge_name)
>     admg.add_edge("Z", "Y", admg.bidirected_edge_name)
>     admg.add_edge("Z", "H", admg.bidirected_edge_name)
> 
>     # not an inducing path, since Y is not a collider and Y is not part of L
>     S = {}
>     L = {}
>     assert not pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]
@adam2392 While I get the point about Y not being in L. But isn't there a path from X - Z - H? In this path, all the nodes obey the rules of inducing path. Unless, what you want is all the nodes on all the possible paths from X -> H have to obey the rules of an inducing path?

adam2392: Ah I see. In this case, X, Y, Z, H is not an inducing path, as discussed.

But if we look at X, Z, H, it is also not an inducing path. Z is not a collider and not part of L.

aryan26roy: I am a bit confused, earlier you said:
> A collider is a collider if there is two arrow endpoints. 

Z has two arrow endpoints pointing towards it. They are due to two bi-directed edges, so do I take it to mean that the two arrow endpoints have to be from directed edges?

adam2392: A collider is a node defined on a path. So it must have two arrow endpoints along the path. Here is something I found online that might help clarify: https://jeangoldinginstitute.blogs.bristol.ac.uk/2019/10/28/a-brief-introduction-to-colliders/

In (X, Z, H) path, Z is not a collider. In (X, Y, Z, H), Z is a collider on the path. Since colliders are defined wrt a path, that is the motivation for my comment on your function `is_collider(...)`, it must be defined wrt a triplet. 

E.g.

```
is_collider(X, Z, H) == False

is_collider(Y, Z, H) == True
```

aryan26roy: Aha! I finally understood what went wrong. Only the nodes on the path count! Looks like I will have to change the implementation slightly. But it can be done easily. Thanks for taking the time to explain this to me.

aryan26roy: One more question @adam2392. You said that any three nodes on the path can be checked if they form a collider, right?

> graph.isDefCollider(a, b, c)

This is what java does. a is the previous node, b is the current node and c is the iterator through all the neighbors of b (except a). It is only checking the collider status of a node w.r.t the node just before and just after it on the path.

adam2392: What is the question? If the collider is defined by 3 nodes? Yes that is a triplet.

aryan26roy: No I get that, what I am asking is, suppose that there is a path:

> A - B - C - D - F - E

When it comes to C, the java implementation is only checking if (B,C,D) form a collider. What if (B,C,F) do too? Will that not count as a collider?

adam2392: Yes but that's a different path? 

There could be a function that returns all possible inducing paths. And another that returns true/false and the first path it finds.

aryan26roy: > # an inducing path, since Y is a collider and is an ancestor of X, H, or S
>     L = {}
>     S = {'A'}
>     assert pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]

@adam2392 Shouldn't this check be :
> assert not pywhy_graphs.inducing_path(admg, "X", "H",L,S)[0]

Or did you mean to add an edge between Y and some other node, not A?

adam2392: Which definition are you looking at? Can you link the paper, or copy/paste the text?

X is an endpoint,
Y is a collider and an ancestor of A, which is in S
Z is a collider and is an ancestor of X
H is an endpoint

therefore (X, Y, Z, H) is an inducing path relative to L and S.

aryan26roy: I read the definition wrong!

>and every collider on p is an ancestor of either X , Y , or a member of S

I thought this meant, every collider has to be an ancestor of X or Y, or a member of S.
But actually it has to be an ancestor of X, Y or a member of S.

aryan26roy: @adam2392 by the same definition, in this example:


>     an inducing path, since Y is a collider and is an ancestor of X, H, or S
>     L = {}
>     S = {'Y'}
>     assert pywhy_graphs.inducing_path(admg, "X", "H", L, S)[0]

Isn't why a collider but not an ancestor of any element in S or X or Y? (It is itself in S)

adam2392: Read the preliminaries section in Zhang 2008 (https://www.sciencedirect.com/science/article/pii/S0004370208001008). By convention A is an ancestor of A.

aryan26roy: Ah. That does make sense. I kind of assumed this was the case and made the modification according to it in the last commit. All the test cases you provided pass. Do you want to take a look at the implementation again?

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,212 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    -------
+    out : set
+        The parents of the provided node.
+    """
+    directed_parents = set(G.sub_directed_graph().predecessors(node))
+
+    return directed_parents
+
+
+def _bidirected_sub_graph_parents(G, node):
```
adam2392, line 359: I wouldn't call these parents. Maybe just say bidirected_sub_graph_neighbors

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,212 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    """Finds the parents of a node in the bidirected subgraph.
+
+    Parameters
+    ----------
+    G : Graph
+        The graph.
+    node : node label
+        The node for which we have to find the parents.
```
adam2392, line 367: Not parents by definition since parents are a uni-directional edge.

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,212 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+
+def _is_collider(G, prev_node, cur_node, node):
+    """Checks if the given node is a collider or not.
+
+    Parameters
+    ----------
+    G : graph
+        The graph.
+    node : node
+        The node to be checked.
```
adam2392, line 390: Update docstring based on function

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,212 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+
+    Returns
+    -------
+    iscollider : bool
+        Bool is set true if the node is a collider, false otherwise.
+    """
+    parents = _directed_sub_graph_parents(G, cur_node)
+    parents = parents.union(_bidirected_sub_graph_parents(G, cur_node))
+
+    if len(parents) == 2 and parents == {prev_node, node}:
```
adam2392, line 400: The number of parents does not need to be 2.

X -> Y <- Z with A -> Y

(X, Y, Z) is still a collider at Y.

adam2392 (changes requested):
Some of the base functions are not correct, so I stopped my review early. I'm surprised the `len(parents)==2` did not raise anymore errors. 

Perhaps you want to add a test case where that specific line causes failure and then you can modify the code afterwards to patch this.

pywhy_graphs/algorithms/generic.py
aryan26roy: Yeah makes sense. I will add a test for it then patch it.

aryan26roy:
@adam2392 I have incorporated all the suggestions you gave in previous reviews except the simplifying the checks one(not sure how I would simplify this further, it looks like a neat abstraction to me). Can you take a look?

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,210 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    ----------
+    G : graph
+        The graph.
+    prev_node : node
+        The previous node in the path.
+    cur_node : node
+        The node to be checked.
+    next_node:
+        The next node in the path.
+    Returns
```
adam2392, line 395: Add spacing between the sections to improve readability.

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,210 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    path : Tuple[bool, path]
+        A tuple containing a bool and a path if the bool is true, an empty list otherwise.
+
+    Notes
+    -----
+    An inducing path intuitively is a path between two non-adjacent nodes that
+    cannot be d-separated. Therefore, the path is always "active" regardless of
+    what variables we condition on. L contains all the non-colliders, these nodes
+    are ignored along the path. S contains all the colliders, these nodes and their
+    ancestors are conditioned upon.
```
adam2392, line 499:
```suggestion
    are ignored along the path. S contains nodes that are always conditioned on (hence if the ancestors of colliders are in S, then those collider paths are always "active").
```

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,210 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+
+    xanc = _directed_sub_graph_parents(G, node_x)
+    yanc = _directed_sub_graph_parents(G, node_y)
+
+    xyancestors = xanc.union(yanc)
+
+    check_ancestors = set()
+
+    for elem in S:
+        check_ancestors = check_ancestors.union(_directed_sub_graph_parents(G, elem))
```
adam2392, line 525: This only gets parents, so:

```
X -> Z <- Y, A <- B <- Z

with 
S = {'A'}

would fail since X has an inducing path to Y with respect to L={} and S={'A'}
```

adam2392 (commented):
Some basic issues that don't match w/ the definition

pywhy_graphs/algorithms/generic.py
aryan26roy: Oh wait, so an ancestor of any node is a node that has a directed path to the said node?

adam2392: Conditioning on a descendant of a collider opens the collider

aryan26roy:
@adam2392 I made the suggested changes and some more.

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,236 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+
+    for elem in children:
+        if elem in visited:
+            continue
+
+        else:
+            if (
+                _is_collider(G, prev_node, cur_node, elem)
+                and (cur_node not in check_ancestors)
+                and (cur_node not in S)
```
adam2392, line 473: Is `S` necessary? Perhaps we can simplify by removing S since ancestors should include S

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,236 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+
+    if cur_node is node_y:
+        return (True, [node_y])
+
+    for elem in children:
+        if elem in visited:
+            continue
+
+        else:
+            if (
```
adam2392, line 470:
```suggestion
# if the current node is a collider, check that...
            if (
```

Add in-line comments to make it clear what this complex if/else sequence is doing

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,236 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+
+    xyancestors = xanc.union(yanc)
+
+    check_ancestors = set()
+
+    for elem in S:
+        check_ancestors = check_ancestors.union(_directed_sub_graph_ancestors(G, elem))
+
+    check_ancestors = xyancestors.union(check_ancestors)
+    children = G.neighbors(node_x)
```
adam2392, line 554: I don't think these are necessarily children?

pywhy_graphs/algorithms/tests/test_generic.py
```diff
@@ -39,3 +40,193 @@ def test_convert_to_latent_confounder(graph_func):
...
+    admg.add_edge("B", "A", admg.directed_edge_name)
+    admg.add_edge("B", "Y", admg.directed_edge_name)
+
+    L = {"X", "B"}
+    S = {"A"}
+
+    assert pywhy_graphs.inducing_path(admg, "Z", "Y", L, S)[0]
+
+
+def test_ancestors_bug_inducing_path():
```
adam2392, line 221: What is the bug?

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,236 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+        Nodes that are ignored on the path. Defaults to an empty set. See Notes for details.
+    S:  set
+        Nodes that are always conditioned on. Defaults to an empty set. See Notes for details.
+
+    Returns
+    -------
+    path : Tuple[bool, path]
+        A tuple containing a bool and a path if the bool is true, an empty list otherwise.
+
+    Notes
```
adam2392, line 518: Can I also trouble you to write a minimal working example? e.g. see networkx/numpy/scikit-learn documentation

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,236 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    parents = _directed_sub_graph_parents(G, cur_node)
+    parents = parents.union(_bidirected_sub_graph_neighbors(G, cur_node))
+
+    if prev_node in parents and next_node in parents:
+        return True
+
+    return False
+
+
+def _recursive_path(G, node_x, node_y, L, S, visited, check_ancestors, cur_node, prev_node):
```
adam2392, line 428: Can you add types for mypy for everything except G?

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,236 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+    L : set
+        Set containing all the non-colliders.
+    S : set
+        Set containing all the colliders.
+    visited : set
+        Set containing all the nodes already visited.
+    check_ancestors : set
+        Set containing all the ancestors to be checked against.
+    cur_node : node
+        The current node.
```
adam2392, line 450: Missing documentation on prev_node

adam2392 (changes requested):
Looking better! Thanks for your hard work! Do you mind cleaning up the PR a bit? As of now some of the variable names, function names, etc. are a bit incongruent.

It makes it harder to read and debug on the fly (for me and future developers and future you).

pywhy_graphs/algorithms/generic.py
aryan26roy: Last I remember the nx.ancestors function doesn't count the same node as it's ancestor. But I can take a look and let you know if I can do away with it.

aryan26roy:
Yeah, patching the bugs has wreaked havoc on my naming system. Will make everything better in my next pass.

pywhy_graphs/algorithms/generic.py
```diff
@@ -425,7 +425,17 @@ def _is_collider(G, prev_node, cur_node, next_node):
     return False
 
 
-def _recursive_path(G, node_x, node_y, L, S, visited, check_ancestors, cur_node, prev_node):
+def _shortest_valid_path(
+    G,
+    node_x: Node,
+    node_y: Node,
+    L: set,
```
adam2392, line 432: Use `Set` instead.

```
from typing import Set
```

pywhy_graphs/algorithms/generic.py
aryan26roy: @adam2392 Where do you want me to add the example? And what format should it be in? Should it be a Python code with comments explaining what's happening or an article with Python snippets?

aryan26roy: Can I add the example to the plot_mixed_edge_graph.ipynb? Would that be automatically included in the docs?

adam2392: You can add one in the examples directory as a Python script. 

You can see some of the patterns used for rendering the example and how it's done in the other examples. 

We basically use Sphinx to render this into a notebook which is ran on the CI pipelines.

aryan26roy: Alright, let me try and write a first draft.

aryan26roy:
@adam2392 The docs are not being built correctly, what am I doing wrong?

pywhy_graphs/algorithms/generic.py
```diff
@@ -333,3 +334,259 @@ def _single_shortest_path_early_stop(G, firstlevel, paths, cutoff, join, valid_p
...
+        A tuple containing a bool and a path if the bool is true, an empty list otherwise.
+
+    Notes
+    -----
+    An inducing path intuitively is a path between two non-adjacent nodes that
+    cannot be d-separated. Therefore, the path is always "active" regardless of
+    what variables we condition on. L contains all the non-colliders, these nodes
+    are ignored along the path. S contains nodes that are always conditioned on
+    (hence if the ancestors of colliders are in S, then those collider
+    paths are always "active").
```
adam2392, line 546: You need a References section at the end here. See some other docstrings.

adam2392 (commented):
The CI error comes cuz there isn't a references section in the docstring of inducing path.

pywhy_graphs/algorithms/generic.py
aryan26roy: @adam2392 I added a references section and the build is still failing.

adam2392: Hmm weird. Perhaps try debugging locally. 

I can take a look this week / tmr.

aryan26roy: My local build has been giving me errors for a while. I will try debugging it. Another thing is the networkx tests are failing because apparently the support for Python 3.8 has been deprecated.

aryan26roy:
@adam2392 do you want me to fix the networkx CI issue in another PR? I can bump up the CI from [3.8, 3.10] to [3.9,3.10].

pywhy_graphs/algorithms/generic.py
adam2392: It turns out you cannot have a `:footcite:`...`` pattern inside the header of the docstring, so I moved it to the description.

examples/inducingpath/inducing_path.py
adam2392, line 1: Perhaps we can just use one graph to illustrate the point. Can we use figure 2 from the paper https://arxiv.org/pdf/1104.5617.pdf

Then proceed by:
1. drawing the figure and describing the inducing path and figure
2. By default adjacent nodes have a trivial inducing path (the edge between them)
3. show that inducing paths among non-adjacent nodes relative to L={L1, L2}, S={} is only possible in the figure between X1 and X5. 
4. There is no inducing path relative to L and S between (X1, X6) and (X2, X4)
5. By adding X6 to S, then there is now an inducing path relative to L and S among (X2, X4), (X1, X3), (X5, X3)

This walks through the core concepts we've walked through on this PR on a graph that is in a well-known publication and then describes some of the intricacies of the hyperparameters (L and S).

adam2392: If you would like to tackle this in another PR, we can merge this one first and you can just remove the `inducing_path.py` file for now.

aryan26roy: Sure, that sounds good.

aryan26roy:
@adam2392 I have removed the example file. Do you want to merge now?

adam2392:
Thanks @aryan26roy !

PR #83, opened 2023-06-21
[DOCS] Add an example for the Inducing Path function

### LINKED ISSUE #81: Update `intro` folder of Sphinx `examples/` to show intuition and concepts of inducing path
As a follow-up to #78. 

Perhaps we can just use one graph to illustrate the point. Can we use figure 2 from the paper https://arxiv.org/pdf/1104.5617.pdf

Then proceed by:
1. drawing the figure and describing the inducing path and figure
2. By default adjacent nodes have a trivial inducing path (the edge between them)
3. show that inducing paths among non-adjacent nodes relative to L={L1, L2}, S={} is only possible in the figure between X1 and X5. 
4. There is no inducing path relative to L and S between (X1, X6) and (X2, X4)
5. By adding X6 to S, then there is now an inducing path relative to L and S among (X2, X4), (X1, X3), (X5, X3)

This walks through the core concepts we've walked through on this PR on a graph that is in a well-known publication and then describes some of the intricacies of the hyperparameters (L and S).

_Originally posted by @adam2392 in https://github.com/py-why/pywhy-graphs/pull/78#discussion_r1235749455_
            
cc: @aryan26roy may be interested :)

aryan26roy:
Closes #81 

- Add an example in examples/intro

## Before submitting

## After submitting

### DIFF
--- a/examples/intro/inducing_path.py
+++ b/examples/intro/inducing_path.py
@@ -63,8 +63,8 @@ print(pywhy_graphs.inducing_path(G, "X1", "X4", L, S))
 print(pywhy_graphs.inducing_path(G, "X3", "X2", L, S))
 
 # %%
-# Inducing Path Two Arbitrary Nodes
-# ----------------------------------------------
+# Inducing paths between non-adjacent nodes
+# ---------------------------------------------
 # Given the definition of an inducing path, we need to satisfy all
 # requirements for the function to return True. Adding the latent
 # variables to L is not enough for the pair [X1,X5]
--- a/examples/intro/inducing_path.py
+++ b/examples/intro/inducing_path.py
@@ -102,7 +102,7 @@ print(pywhy_graphs.inducing_path(G, "X1", "X5", L, S))
 # this returns False
 print(pywhy_graphs.inducing_path(G, "X1", "X3", L, S))
 
-# We need to add X6, which will open up paths
+# If we add X6, we will open up certain collider paths
 # including all the collider ancestors of X6
 # in this case that node is X2.
 
--- a/examples/intro/inducing_path.py
+++ b/examples/intro/inducing_path.py
@@ -44,6 +44,7 @@ G.add_edge("X6", "X4", G.circle_edge_name)
 G.add_edge("X6", "X5", G.circle_edge_name)
 
 
+# this is the Figure 2(a) in the paper as we see.
 dot_graph = draw(G)
 dot_graph.render(outfile="pag.png", view=True)
 
--- a/examples/intro/inducing_path.py
+++ b/examples/intro/inducing_path.py
@@ -30,8 +30,10 @@ G.add_edge("X3", "X6", G.directed_edge_name)
 G.add_edge("X3", "X4", G.directed_edge_name)
 G.add_edge("X3", "X2", G.directed_edge_name)
 G.add_edge("X5", "X6", G.directed_edge_name)
-G.add_edge("X2", "X1", G.bidirected_edge_name)
-G.add_edge("X4", "X5", G.bidirected_edge_name)
+G.add_edge("L2", "X4", G.directed_edge_name)
+G.add_edge("L2", "X5", G.directed_edge_name)
+G.add_edge("L1", "X1", G.directed_edge_name)
+G.add_edge("L1", "X2", G.directed_edge_name)
 G.add_edge("X2", "X3", G.circle_edge_name)
 G.add_edge("X4", "X3", G.circle_edge_name)
 G.add_edge("X6", "X4", G.circle_edge_name)
@@ -41,10 +43,46 @@ G.add_edge("X6", "X5", G.circle_edge_name)
 dot_graph = draw(G)
 dot_graph.render(outfile="pag.png", view=True)
 
-# X2 is the only collider on the path
+# All adjacent nodes have a trivial inducing path
+# which is the edge between them
 L = {}
-S = {"X2"}
+S = {}
+
+# All such tests will return True.
+print(pywhy_graphs.inducing_path(G, "X1", "X4", L, S))
+print(pywhy_graphs.inducing_path(G, "X3", "X2", L, S))
 
 
-# returns true
+# Including all the 'L' nodes is not enough to
+# open up an inducing path between X1 and X5
+L = {"L1", "L2"}
+S = {}
+
+
+# returns False
+print(pywhy_graphs.inducing_path(G, "X1", "X5", L, S))
+
+
+# However, if we add X3, a non-collider on the path
+# from X1 to X5 to L, we open up an inducing path.
+
+
+L = {"L1", "L2", "X3"}
+S = {}
+
+# now it returns True
+print(pywhy_graphs.inducing_path(G, "X1", "X5", L, S))
+
+# Even now, some inducing paths are not opened.
+# For example, the path between X1 and X3 is not available
+
+# this returns False
+print(pywhy_graphs.inducing_path(G, "X1", "X3", L, S))
+
+# We need add X6, a collider on the path, to S
+
+L = {"L1", "L2", "X3"}
+S = {"X6"}
+
+# now it returns True
 print(pywhy_graphs.inducing_path(G, "X1", "X3", L, S))
--- a/examples/intro/inducing_path.py
+++ b/examples/intro/inducing_path.py
@@ -66,11 +66,14 @@ print(pywhy_graphs.inducing_path(G, "X3", "X2", L, S))
 # ------------------------------------------
 # Given the definition of an inducing path, we need to satisfy all
 # requirements for the function to return True. Adding the latent
-# variables to L is not enough for the pair [X1,X5]. As we see in Figure 2(c) in :footcite:`Colombo2012`,
-# (X1, X5) are not adjacent in the final skeleton of the equivalence class, which makes sense because a
-# MAG is an equivalence class of a DAG and contains an edge among two nodes if i) the two nodes are
-# adjacent in the DAG, or ii) the two nodes have a primitive inducing path between them. Since there
-# is no adjacency among (X1, X5) in the final skeleton, there is no primitive inducing path between them.
+# variables to L is not enough for the pair [X1,X5]. As we see in
+# Figure 2(c) in :footcite:`Colombo2012`,  (X1, X5) are not adjacent
+# in the final skeleton of the equivalence class, which makes sense
+# because a MAG is an equivalence class of a DAG and contains an
+# edge among two nodes if i) the two nodes are adjacent in the DAG,
+# or ii) the two nodes have a primitive inducing path between them.
+# Since there is no adjacency among (X1, X5) in the final skeleton,
+# there is no primitive inducing path between them.
 
 L = {"L1", "L2"}
 S = {}
@@ -117,8 +120,6 @@ S = {"X6"}
 print(pywhy_graphs.inducing_path(G, "X1", "X3", L, S))
 
 
-"""
-References
-----------
-.. footbibliography::
-"""
+# References
+# ----------
+# .. footbibliography::
--- a/examples/intro/inducing_path.py
+++ b/examples/intro/inducing_path.py
@@ -22,7 +22,7 @@ from pywhy_graphs.viz import draw
 # The Graph
 # ----------------------------------------------
 # To illustrate the workings of the inducing path algorithm, we will
-# construct the causal graph from figure 2 of https://arxiv.org/pdf/1104.5617.pdf.
+# construct the causal graph from figure 2 of :footcite:`Colombo2012`.
 
 
 G = PAG()
--- a/examples/intro/inducing_path.py
+++ b/examples/intro/inducing_path.py
@@ -18,8 +18,9 @@ import pywhy_graphs
 from pywhy_graphs import PAG
 from pywhy_graphs.viz import draw
 
-# construct a causal graph that will result in
-# X <- Y <-> Z <-> H; Z -> X
+# construct the causal graph from figure 2 of
+# https://arxiv.org/pdf/1104.5617.pdf
+
 G = PAG()
 G.add_edge("X4", "X1", G.directed_edge_name)
 G.add_edge("X2", "X5", G.directed_edge_name)
@@ -36,6 +37,9 @@ G.add_edge("X4", "X3", G.circle_edge_name)
 G.add_edge("X6", "X4", G.circle_edge_name)
 G.add_edge("X6", "X5", G.circle_edge_name)
 
+
+dot_graph = draw(G)
+dot_graph.render(outfile="pag.png", view=True)
 
 # X2 is the only collider on the path
 L = {}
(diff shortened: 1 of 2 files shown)


aryan26roy:
@adam2392 How do you want me to proceed with this? Should I include scenarios where there is no path as well? And are the comments like this ok?

adam2392:
> @adam2392 How do you want me to proceed with this? Should I include scenarios where there is no path as well? And are the comments like this ok?

The goal is to educate the user and serve as a quick reminder for yourself as a dev in case you need to review something.

I would proceed in sections:

1. setup the data and graph
2. draw the graph, perhaps comment that there is only 1 inducing path relative to L and S. This one inducing path is why the MAG in Figure 2 of the paper contains an extra adjacency
3. Run the inducing path algorithm and actually verify this 1 inducing path and verify a few non-inducing path examples as well
4. Show and explain what happens when we include the node into set S

Lmk if that makes sense

aryan26roy:
@adam2392 What do you think?

adam2392:
Yes I think this is a good start and direction!

Can you take a look at the other examples and use "# %%" to break up into sections where it makes sense?

It makes the final rendering easier to read.

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,111 @@
+"""
+======================================================
+An introduction to Inducing Paths and how to find them
+======================================================
+
+A path p is called an ``inducing path`` relateve to <L,S>
+on an ancestral graph, where every non-endpoint vertex on p
+is either in L or a collider, and every collider on p is an ancestor
+of either X , Y or a member of S.
```
adam2392, line 9: More details on inducing paths can be found at :footcite:`Zhang2008`.

Then add the .. footbibliography:: at the bottom, so the references render.

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,111 @@
...
+
+import pywhy_graphs
+from pywhy_graphs import PAG
+from pywhy_graphs.viz import draw
+
+# %%
+# The Graph
+# ----------------------------------------------
+# To illustrate the workings of the inducing path algorithm, we will
+# construct the causal graph from figure 2 of https://arxiv.org/pdf/1104.5617.pdf.
```
adam2392, line 25:
```suggestion
# construct the causal graph from figure 2 of :footcite:`Colombo2012`.
```

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,111 @@
...
+we condition on.
+"""
+
+import pywhy_graphs
+from pywhy_graphs import PAG
+from pywhy_graphs.viz import draw
+
+# %%
+# The Graph
+# ----------------------------------------------
```
adam2392, line 23: Too many '-' characters.

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,111 @@
...
+d-seperated, making it active regardless of what variables
+we condition on.
+"""
+
+import pywhy_graphs
+from pywhy_graphs import PAG
+from pywhy_graphs.viz import draw
+
+# %%
+# The Graph
```
adam2392, line 22:
```suggestion
# Construct an example graph
```

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,111 @@
...
+G.add_edge("L2", "X5", G.directed_edge_name)
+G.add_edge("L1", "X1", G.directed_edge_name)
+G.add_edge("L1", "X2", G.directed_edge_name)
+G.add_edge("X2", "X3", G.circle_edge_name)
+G.add_edge("X4", "X3", G.circle_edge_name)
+G.add_edge("X6", "X4", G.circle_edge_name)
+G.add_edge("X6", "X5", G.circle_edge_name)
+
+
+dot_graph = draw(G)
```
adam2392, line 47:
```suggestion
# this is the Figure 2(a) in the paper as we see.
dot_graph = draw(G)
```

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,111 @@
...
+from pywhy_graphs.viz import draw
+
+# %%
+# The Graph
+# ----------------------------------------------
+# To illustrate the workings of the inducing path algorithm, we will
+# construct the causal graph from figure 2 of https://arxiv.org/pdf/1104.5617.pdf.
+
+
+G = PAG()
```
adam2392, line 28: You want to draw the graph, not the PAG?

So there shouldn't be any circle edges since there is no such thing as an inducing path for PAGs, or CPDAGs.

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,111 @@
...
+G.add_edge("X6", "X4", G.circle_edge_name)
+G.add_edge("X6", "X5", G.circle_edge_name)
+
+
+dot_graph = draw(G)
+dot_graph.render(outfile="pag.png", view=True)
+
+
+# %%
+# Inducing Path Between Adjacent Nodes
```
adam2392, line 52:
```suggestion
# Adjacent nodes trivially have an inducing path
```

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,111 @@
...
+L = {}
+S = {}
+
+# All such tests will return True.
+print(pywhy_graphs.inducing_path(G, "X1", "X4", L, S))
+print(pywhy_graphs.inducing_path(G, "X3", "X2", L, S))
+
+# %%
+# Inducing Path Two Arbitrary Nodes
+# ----------------------------------------------
```
adam2392, line 66:
```suggestion
# Inducing paths between non-adjacent nodes
# ---------------------------------------------
```

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,111 @@
...
+S = {}
+
+# now it returns True
+print(pywhy_graphs.inducing_path(G, "X1", "X5", L, S))
+
+
+# %%
+# The Role of Colliders
+# ----------------------------------------------
+# Adding Colliders to the set S has a downstream effect.
```
adam2392, line 93:
```suggestion
# Adding colliders to the set S has a downstream effect.
```

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,111 @@
...
+# Since adding the collider opens up all the inducing paths containing
+# the collider ancestors of that node.
```
adam2392, line 95:
```suggestion
# Conditioning on a collider, or descendant of a collider opens up that collider path.
# For example, we will add the node 'X6' to the set ``S``. This will open up the collider
# path ``(X1, X2, X3)``, since 'X2' is an ancestor of 'X6'.
```

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,111 @@
...
+
+# Even now, some inducing paths are not opened.
+# For example, the path between X1 and X3 is not available
+
+# this returns False
+print(pywhy_graphs.inducing_path(G, "X1", "X3", L, S))
+
+# We need to add X6, which will open up paths
```
adam2392, line 103:
```suggestion
# If we add X6, we will open up certain collider paths
```

adam2392 (changes requested):
There are a few comments I left. The most major one is that I'm surprised the code ran if you used a PAG with circle-edges. 

An inducing path definition only mentions directed and bidirected edges, so for any graph with other kinds of edges (e.g. CPDAG or PAG) it is not defined what it is.

aryan26roy:
@adam2392 can you take a look now?

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,116 @@
...
+
+# %%
+# The role of colliders
+# ----------------------
+# Adding colliders to the set S has a downstream effect.
+# Conditioning on a collider, or descendant of a collider opens up that collider path.
+# For example, we will add the node 'X6' to the set ``S``. This will open up the collider
+# path ``(X1, X2, X3)``, since 'X2' is an ancestor of 'X6'.
+
+# Even now, some inducing paths are not opened.
```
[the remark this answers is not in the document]
aryan26roy: Sure.

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,116 @@
...
+G.add_edge("X5", "X6", G.directed_edge_name)
+G.add_edge("L2", "X4", G.directed_edge_name)
+G.add_edge("L2", "X5", G.directed_edge_name)
+G.add_edge("L1", "X1", G.directed_edge_name)
+G.add_edge("L1", "X2", G.directed_edge_name)
+
+
+# this is the Figure 2(a) in the paper as we see.
+dot_graph = draw(G)
+dot_graph.render(outfile="pag.png", view=True)
```
adam2392, line 51: Maybe let's rename it to `graph.png` since pag is misleading

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,116 @@
...
+# All such tests will return True.
+print(pywhy_graphs.inducing_path(G, "X1", "X4", L, S))
+print(pywhy_graphs.inducing_path(G, "X3", "X2", L, S))
+
+# %%
+# Inducing paths between non-adjacent nodes
+# ------------------------------------------
+# Given the definition of an inducing path, we need to satisfy all
+# requirements for the function to return True. Adding the latent
+# variables to L is not enough for the pair [X1,X5]
```
adam2392, line 72:
```suggestion
# variables to L is not enough for the pair [X1,X5]. As we see in Figure 2(c) in :footcite:`Colombo2012`,
# (X1, X5) are not adjacent in the final skeleton of the equivalence class, which makes sense because a
# MAG is an equivalence class of a DAG and contains an edge among two nodes if i) the two nodes are
# adjacent in the DAG, or ii) the two nodes have a primitive inducing path between them. Since there
# is no adjacency among (X1, X5) in the final skeleton, there is no primitive inducing path between them.
```

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,120 @@
...
+
+In other words, it is a path between two nodes that cannot be
+d-seperated, making it active regardless of what variables
+we condition on.
+
+More details on inducing paths can be found at :footcite:Zhang2008.
+
+References
+----------
+.. footbibliography::
```
adam2392, line 20: This doesn't render properly. I think you might need to add this to the very bottom of the file.

https://output.circle-artifacts.com/output/job/9f6f31fd-869c-41c9-83ce-c66a5107c2d6/artifacts/0/dev/auto_examples/intro/inducing_path.html#sphx-glr-auto-examples-intro-inducing-path-py

aryan26roy:
> Can you add a test in `inducing_path` that a graph should only have directed and/or bidirected edges? Otw, and inducing path doesn't make sense to run.

Do you want me to add it to this PR or create a new issue and a new PR?

adam2392:
You can just add it here since it's just a line just checking the types of edges present.

aryan26roy:
@adam2392 there's one problem. There is no graph class that is only has directed and bidirected edges. Even ADMG can have undirected edges. Do you want me to just check if the argument is an instance of PAG or CPDAG instead?

adam2392:
> @adam2392 there's one problem. There is no graph class that is only has directed and bidirected edges. Even ADMG can have undirected edges. Do you want me to just check if the argument is an instance of PAG or CPDAG instead?

I wouldn't check the class as we are going to eventually try to refactor that so there is no explicit PAG/CPDAG class.

Instead, I would check the edge types for right now 'directed', 'bidirected' keywords.

aryan26roy:
> Instead, I would check the edge types for right now 'directed', 'bidirected' keywords.

The problem is PAG would pass this test. It can have both of these edge types.

adam2392:
> > Instead, I would check the edge types for right now 'directed', 'bidirected' keywords.
> 
> The problem is PAG would pass this test. It can have both of these edge types.

We don't want any other edge. It can only have these types of edges.

aryan26roy:
@adam2392 Unless you're ok with instances of PAG class being valid as long as they only have directed and bidirected edges. Otherwise there is no way to differentiate between the different types.

aryan26roy:
@adam2392 I added the check. Also, I can't get the references to render, can you take a look at both?

examples/intro/inducing_path.py
```diff
@@ -0,0 +1,125 @@
...
+on an ancestral graph, where every non-endpoint vertex on p
+is either in L or a collider, and every collider on p is an ancestor
+of either X , Y or a member of S.
+
+
+In other words, it is a path between two nodes that cannot be
+d-seperated, making it active regardless of what variables
+we condition on.
+
+More details on inducing paths can be found at :footcite:Zhang2008.
```
adam2392, line 16:
```suggestion
More details on inducing paths can be found at :footcite:`Zhang2008`.
```

adam2392 (commented):
I think the references did render(?)

https://output.circle-artifacts.com/output/job/6ee4db77-2b61-47f3-af28-eac233b61576/artifacts/0/dev/auto_examples/intro/inducing_path.html#sphx-glr-auto-examples-intro-inducing-path-py

aryan26roy:
@adam2392 The last run didn't make any documentation artifacts?

adam2392:
Are you able to see the status of the circle CI docs build? The artifacts are listed there.

pywhy_graphs/algorithms/generic.py
```diff
@@ -565,6 +565,13 @@ def inducing_path(G, node_x: Node, node_y: Node, L: Set = None, S: Set = None):
     if node_x == node_y:
         raise ValueError("The source and destination nodes are the same.")
 
+    edges = G.edges()
+
+    for elem in edges.keys():
+        if elem not in {"directed", "bidirected"}:
+            if len(edges[elem]) != 0:
+                raise ValueError("Inducing Path is not defined for this graph.")
```
adam2392, line 573: I would instead do something like:

```Python
if any(edge_type not in {'directed', 'bidirected'} for edge_type in G.edge_types):
     raise...
```

See: https://www.pywhy.org/pywhy-graphs/dev/generated/pywhy_graphs.networkx.MixedEdgeGraph.html#pywhy_graphs.networkx.MixedEdgeGraph

aryan26roy: @adam2392 If I use this test all of the ADMG classes are raising errors as well.

adam2392: Oh okay I see. Can you add a comment marking this as something we should fix in the future `# XXX: fix this when graph classes are refactored to only check for directed/bidirected edge types`?

adam2392 (changes requested):
Some minor changes

pywhy_graphs/algorithms/generic.py
```diff
@@ -565,6 +565,13 @@ def inducing_path(G, node_x: Node, node_y: Node, L: Set = None, S: Set = None):
     if node_x == node_y:
         raise ValueError("The source and destination nodes are the same.")
 
+    edges = G.edges()
+
+    for elem in edges.keys():
```
adam2392, line 570:
```suggestion
    # XXX: fix this when graph classes are refactored to only check for directed/bidirected edge types
    for elem in edges.keys():
```

adam2392 (approved):

adam2392:
Thanks @aryan26roy !

PR #87, opened 2023-07-17
[ENH] Add the ability to convert an MAG to a PAG.

### LINKED ISSUE #85: Convert MAG to PAG
A MAG is a mixed-edge graph with directed and bidirected edges and has the 'maximality' property. 

This is a sub-procedure of https://github.com/py-why/pywhy-graphs/pull/73 and will enable that to function properly.

Implement a function `mag_to_pag`, which takes in a valid MAG and outputs the corresponding PAG.

**Additional context**
@aryan26roy if you're still interested, this can be the next step before revisiting #73.

aryan26roy:
Closes #85

- Adds a function to convert an MAG to PAG.

## Before submitting

## After submitting


aryan26roy:
@adam2392 I will look for the exact algorithm and add it to the PR description.

aryan26roy:
@adam2392 I can't find an implementation in the tetrad code base. Do you know where might I find it?

aryan26roy:
@adam2392 for converting an MAG to PAG, do you want me to implement the FCI rules here? Or can we use the DagToPag class from dodiscover?

adam2392:
> @adam2392 for converting an MAG to PAG, do you want me to implement the FCI rules here? Or can we use the DagToPag class from dodiscover?

Hmm good question. So the issue is that some of the rules rely on knowledge of the separating set, which basically requires re-running FCI with the graph as an oracle for the Conditional independence (CI) test.

For now, you can just import dodiscover and run FCI. However, this might result in a bit of an annoyance as dodiscover API is relatively unstable. I am disinclined to add this to the pywhy-graphs codebase as a result.

Sorry for bouncing you around, but perhaps we can revisit the PAG_to_MAG algorithm? https://github.com/py-why/pywhy-graphs/issues/67#issue-1632126474

To perform the check on a PAG, it seems you have:

- [x] check for some validity that it is a PAG (e.g. circle edges); you should be able to do this pretty easily
- [ ] convert the PAG to a MAG using the theorem linked in the issue in Zhang's 2008 paper (we don't have a MAG class, but we can just use an ADMG class for now)
- [x] check the validity of the MAG
- [ ] convert the MAG back to a PAG
- [x] check if PAG is same as the PAG fed in; easy to do.

The MAG to a PAG is technically implemented in dodiscover via FCI with the graph as an oracle. I will need to think about how to best add this in. I would now focus on the PAG to MAG case. To setup the unit-tests, we can use dodiscover to convert DAGs to PAGs.

adam2392:
Basically tldr: @aryan26roy let's start a new PR on doing PAG to MAG. In the unit-tests you can use dodiscover for now if you need to construct a valid PAG from a DAG. Alternatively, you can use Tetrad GUI to come up with some random PAGs.

aryan26roy:
> For now, you can just import dodiscover and run FCI. However, this might result in a bit of an annoyance as dodiscover API is relatively unstable. I am disinclined to add this to the pywhy-graphs codebase as a result.

Should I go ahead with this @adam2392 ?

aryan26roy:
> This meaning adding FCI rules to the codebase?

No, I meant importing dodiscover in pywhy-graphs. I was assuming that the API would have stabilised by now.

adam2392:
> > This meaning adding FCI rules to the codebase?
> 
> No, I meant importing dodiscover in pywhy-graphs. I was assuming that the API would have stabilised by now.

It is not yet, so you can just add the FCI rules temporarily here to help convert a MAG to a PAG

aryan26roy:
@adam2392 iirc, the FCI algorithm only converts a DAG to a PAG. Won't I need to convert the MAG to a DAG first?

adam2392:
> @adam2392 iirc, the FCI algorithm only converts a DAG to a PAG. Won't I need to convert the MAG to a DAG first?

nope. I would remind yourself what the difference in skeleton structure between a DAG and a MAG ;)

aryan26roy:
> nope. I would remind yourself what the difference in skeleton structure between a DAG and a MAG ;)

It's just that an MAG has more kinds of edges, right?

adam2392:
What invariances does an equivalence class like a MAG encode? Do those invariances mean anything probabilistically and graphically? How does the skeleton of the PAG compare to the MAG and DAG? Finally, does it matter for FCI what is fed in? I'll let you explore the rest as I think the answer will be enlightening!

Feel free to ping me about this PR when it's ready to review.

adam2392:
I resolved CI issues in #99 . Feel free to merge/rebase changes

aryan26roy:
@adam2392 I will read up on this and then implement the algorithms. Thanks for the help!

adam2392:
> @adam2392 I will read up on this and then implement the algorithms. Thanks for the help!

The FCI rules can prolly be copied over as a refactored version from dodiscover. But yes agreed that reading up on this will help you understand how to test the MAG -> PAG properly and later PRs

adam2392:
Sure

aryan26roy:
@adam2392 The FCI implementation in dodiscover uses a lot of utilities from the dodiscover library. It is also a sub-class of another class called BaseConstraintDiscovery. Should I implement all of these things in pywhy-graphs or simply import them from dodiscover?

adam2392:
> Hmm okay I see, then perhaps let's forego this. It's too much work to replicate the code here and not worth it since it's already in dodiscover.
> 
> We need this I guess for #67 ?
> 
> I think at that step, we can just use dodiscover code for now.

@aryan26roy just checking in here. Just wanted to make sure this seems fine to you?

aryan26roy:
@adam2392 sorry for the late reply, this was the Diwali weekend and I was on vacation. Yes, this seems alright to me. Will use the dodiscover code for now.

PR #93, opened 2023-08-22
[ENH] Add the ability to convert a PAG to MAG

### LINKED ISSUE #92: Convert PAG to MAG
I want to be able to convert a PAG to an MAG.

A function that takes in a PAG and outputs a valid MAG.

**Describe alternatives you've considered**

**Additional context**
Needed for #67

aryan26roy:
Closes #92 

- A function that converts a PAG to an MAG.

## Before submitting

## After submitting


aryan26roy:
@adam2392 I can't find the `DagToPag` class in dodiscover. Could you point me to it's implementation?

adam2392:
> @adam2392 I can't find the `DagToPag` class in dodiscover. Could you point me to it's implementation?

I would just use Tetrad to draw up a few graphs; it has the nice functionality of checking if the PAG is valid (see https://github.com/cmu-phil/tetrad/issues/1556). If you read the orientation rules, by breaking any of them, then you can violate the PAG.

The DAGtoPAG class does not exist in dodiscover. Sorry that was just a  shorthand for saying you can use the FCI algo with a graph as the Oracle CI test.

aryan26roy:
@adam2392 converting a PAG to MAG requires implementing the Meek rules. Is there an implementation already available for it? I remember seeing something like that in some pywhy repo.

adam2392:
What does the PAG to MAG algo look like?

aryan26roy:
I am following this implementation in tetrad: https://github.com/cmu-phil/tetrad/blob/c9ac3d8d900c5917b363055f5d4f04b15068229b/tetrad-lib/src/main/java/edu/cmu/tetrad/search/utils/GraphSearchUtils.java#L411

adam2392:
Does THM 2 in the paper use these? Can you maybe summarize the algorithm here or in a docstring? Idr where the meek rules come in or why.

aryan26roy:
The tetrad implementation does exactly what THM 2 asks in step 1. The step 2 involves removing any unshielded colliders. So I just assumed Meek rules do just that. So now I have two questions:
- What do the Meek rules do exactly?
- Is there a better way of getting rid of unshielded colliders?

adam2392:
You should put this into the PR description for making communication easier to reference. Apologies, but if you can help me see where the confusion lies, it will help in answering your questions.

"Let H be the graph resulting from the following procedure applied to the PAG, P:

(1) orient the circles on `o->` edges in P as tails, and orient the circles on `--o` edges in P as arrowheads (that is, turn all `o->` edges and all `--o` edges into directed edges `->`); and
(2) orient $P^C$ into a DAG with no unshielded colliders.
"

The first part doesn't require anything except changing edges in the graph?
The second part just means you orient the circle-edge subgraph edges. Looking at the tetrad implementation, I see this involves the MEEK rules. So the Meek rules are just R1-R4 of the PC algorithm. 

> The tetrad implementation does exactly what THM 2 asks in step 1. The step 2 involves removing any unshielded colliders. So I just assumed Meek rules do just that. So now I have two questions:
> 
> * What do the Meek rules do exactly?

Read up on the PC algorithm and FCI algorithm. The FCI algorithm is described in the Zhang 2008 paper. A version of the PC algorithm is described in https://arxiv.org/pdf/1302.4972.pdf.

> * Is there a better way of getting rid of unshielded colliders?

Not sure what the question is here.

aryan26roy:
>You should put this into the PR description for making communication easier to reference.
Sure, will do this from next time.

> Not sure what the question is here.

I was wondering if there was a pre-implemented way of getting rid of unshielded colliders in pywhy-graphs?

adam2392:
> I was wondering if there was a pre-implemented way of getting rid of unshielded colliders in pywhy-graphs?

This isn't a "standard algorithm", so not that I'm aware of. But the idea of getting rid of unshielded colliders stems from the fact that unshielded colliders provide conditional independence (CI) constraints. 

DAGs encode CI constraints via d-separation. However, many DAGs can have the same CI constraints. Thus the literature invented MAGs. However, many MAGs can have the same CI constraints as well. Thus the literature invented PAGs. So the algorithm doesn't specify how to do it, just that you should be able to take the Thm2 step 2 and orient randomly as long as there isn't an unshielded collider.

E.g.

`A o-o B o-o C`, with A and C not connected should not get oriented into `A o-> B <-o C`.

adam2392 (commented):
Just left a short comment cuz I had some time during lunch, but lmk when you want me to properly review this.

pywhy_graphs/algorithms/pag.py
```diff
@@ -908,3 +909,272 @@ def _check_ts_node(node):
...
+            # Make i-j into i->j
+            # logger.info(f"R2: Removing edge {i}-{j} to form {i}->{j}.")
+            graph.orient_uncertain_edge(i, j)
+            added_arrows = True
+    return added_arrows
+
+
+def pag_to_mag(graph):
+    """Convert an PAG to an MAG.
+
```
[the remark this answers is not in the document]
aryan26roy: Sure. I will add that.

aryan26roy:
@adam2392 When I generated a MAG - PAG pair, I go the following edges for the MAG:

> C <- A -> B -> D; C -> D <- A; 

Now sometimes, my implementation of the PAG to MAG function produces this exact graph, but sometimes it produces these as well:

> [('A', 'D'), ('A', 'C'), ('C', 'D'), ('B', 'D'), ('B', 'A')]

and 

> [('A', 'D'), ('A', 'B'), ('C', 'D'), ('C', 'A'), ('B', 'D')] 

I remember you said something about there being multiple right MAGs (the MAGs would have to be markov equivalent is what you said, iirc). Are these alternate graphs right? Or is my implementation wrong?

adam2392:
> @adam2392 When I generated a MAG - PAG pair, I go the following edges for the MAG:
> 
> > C <- A -> B -> D; C -> D <- A;
> 
> Now sometimes, my implementation of the PAG to MAG function produces this exact graph, but sometimes it produces these as well:
> 
> > [('A', 'D'), ('A', 'C'), ('C', 'D'), ('B', 'D'), ('B', 'A')]
> 
> and
> 
> > [('A', 'D'), ('A', 'B'), ('C', 'D'), ('C', 'A'), ('B', 'D')]

Aren't these the same skeleton? Just not in the same order?

aryan26roy:
The deference is hard to notice. If you look at the first result, the edge A -> B is B -> A and in the second result, A -> C is the other way, i.e  C -> A.

adam2392:
Zhang has a theorem stating when two MAGS are markov equivalent. You can check those three conditions easily by hand for a small graph.

aryan26roy:
I could not find a theorem but I did find these:

>Proposition 2. Two MAGs over the same set of vertices are Markov equivalent if and only if
(e1) They have the same adjacencies;
(e2) They have the same unshielded colliders;
(e3) If a path p is a discriminating path for a vertex V in both graphs, then V is a collider on the path in one graph if and only if it is a
collider on the path in the other.

and

>Definition 5 (Markov equivalence). Two MAGs G1,G2 (with the same set of vertices) are Markov equivalent if for any three disjoint sets of vertices X,Y,Z, X and Y are m-separated by Z in G1 if and only if X and Y are m-separated by Z in G2.

Is this what you are talking about? If not, can you point me to the exact place where it is defined?

adam2392:
Proposition 1 is the one that you can check

aryan26roy:
Assuming that you meant proposition 2, can you explain what exactly point 3 means in the text?

adam2392:
Discriminating path is a graphical property defined in Zhang I think(?) lmk. 

They should also provide an example. I would take a look and walk thru examples and counter examples. 

There is also an implementation in pywhy-graphs if that helps the intuitive understanding.

aryan26roy:
Alright, from what I see they seem Markov equivalent.

aryan26roy:
How many tests should I add to finish this PR?

adam2392:
> How many tests should I add to finish this PR? It doesn't seem like I will be able to construct test cases to hit every line in this.

You don't have to test every line in the `_apply_meek_rules*`, but you should be able to construct a test-case for each line in `pag_to_mag`. As of right now, there seem to be some CI failures.

adam2392:
To check edge presence you should use the API 'has_edge' rather than check tuple presence which is sensitive to order.

aryan26roy:
> To check edge presence you should use the API 'has_edge' rather than check tuple presence which is sensitive to order.

Ah! I didn't know about this.

aryan26roy:
I was just about to comment that.

aryan26roy:
@adam2392 I just realised, if I turned an DAG to an MAG, isn't it a theorem that the resulting MAG will only have directed edges?

adam2392:
> @adam2392 I just realised, if I turned a DAG to an MAG, isn't it a theorem that the resulting MAG will only have directed edges?

When you say directed edges, you mean including bidirected edges? I believe undirected is also allowed if there is selection bias (i.e. non-empty S set) and latent confounders (i.e. non-empty L set), but rarely used. You can fact-check me there.

aryan26roy:
> > @adam2392 I just realised, if I turned a DAG to an MAG, isn't it a theorem that the resulting MAG will only have directed edges?
> 
> When you say directed edges, you mean including bidirected edges? I believe undirected is also allowed if there is selection bias (i.e. non-empty S set) and latent confounders (i.e. non-empty L set), but rarely used. You can fact-check me there.

You're right. They can contain undirected and bidirected edges. But I wanted to ask a different question and asked this one because of a type. What I wanted to ask was can a PAG converted to MAG have directed and undirected edges. I just looked over the paper and turns out the answer to this is yes as well.

Now what I want to know is, if I have two graphs (pywhy-graphs) how do I copy all the edges from one graph to another?

adam2392:
> You're right. They can contain undirected and bidirected edges. But I wanted to ask a different question and asked this one because of a typo. What I wanted to ask was can a PAG converted to MAG have bidirected and undirected edges. I just looked over the paper and turns out the answer to this is yes as well.
> 
> Now what I want to know is, if I have two graphs (pywhy-graphs) how do I copy all the edges from one graph to another?

Try this: https://www.pywhy.org/pywhy-graphs/dev/generated/pywhy_graphs.networkx.MixedEdgeGraph.html#pywhy_graphs.networkx.MixedEdgeGraph.copy

aryan26roy:
I don't understand. This only returns a copy of the graph, right?

adam2392:
> I don't understand. This only returns a copy of the graph, right?

What are you trying to do?

adam2392:
> I have a graph that I need transfer all the edges from into a new instance of ADMG.

I don't think we have a function for that. You can just loop over edge_types and edges.

aryan26roy:
@adam2392 I um unable to hit a codepath in the PR. It requires that the PAG have a o-- edge. What conditions does the DAG need to full fill for the PAG to have a o-- edge?

adam2392:
These are when one has selection bias. If you read the Zhang FCI paper it will tell you in the orientation rules. 

I think R5-7? 

--o just means it could be an undirected edge in the MAG or it could be a directed edge. But the tail is a true tail. So what does that mean?

aryan26roy:
Not sure I understand what you are trying to explain here. Looking at the R5-7, not sure how I would construct a DAG that results in a PAG with the --o edge. Could you maybe show me one such example?

adam2392:
> Not sure I understand what you are trying to explain here. Looking at the R5-7, not sure how I would construct a DAG that results in a PAG with the --o edge. Could you maybe show me one such example?

I meant, look at R5-7 and see which input conditions can result in `--o`. I would suggest working through some small examples and counter-examples. Working backwards, you can see that those input conditions are a function of other rules. For example:

R6: `If A - u o-* C then orient u o-* C as u -* C`

which means we can obtain a `u -o C` hypothetically. R6 input conditions require a `A - u`. A completely undirected edge can be the result of R5:

```
For each A o-o U, if there is an uncovered (e.g. every triple is unshielded)
        circle path p = <A, B, ..., D, U> such that A, D are not adjacent and U, B are
        not adjacent, then orient A o-o U and every edge on p as undirected.
```

Thus if you find scenario where R5 gives you `A - u o-o C`, then R6 will give you `A - u -o C`.

aryan26roy:
@adam2392 It seems like I am unable to construct the right example. Can you construct one and show me how to do it?

adam2392:
> @adam2392 It seems like I am unable to construct the right example. Can you construct one and show me how to do it?

According to the Zhang completeness paper on page 1880, R5-7 are the only rules used for selection bias. R6 and R7 are only applicable if R5 is ran. R5 is only able to be run run if there is an uncovered circle path. Okay, let's determine when R5 would be applicable. I can just construct the simplest possible example I can think of that is exactly what the rule states:

A -> [S1] <- B -> [S2] <- C -> [S3] <- D; A -> [S4] <- D

would result in 

A o-o B o-o C o-o D; A o-o D

Applying R5 gives you undirected edges. Lmk if there is a mistake

aryan26roy:
> A -> [S1] <- B -> [S2] <- C -> [S3] <- D; A -> [S4] <- D
What are S1, S2 etc. in this example? Selection variables? I ask because there is no way to add selection variables in tetrad when constructing a DAG.

adam2392:
Yes I [] indicates a latent variable in my drawing. 

Ah I see. Yeah I guess tetrad FCI doesn't support selection bias. Dodiscover does tho

adam2392:
Cc: @jaron-lee who implemented the rules in dodiscover. 

does my example seem correct for selection bias?

aryan26roy:
@adam2392 is there an example online that I can follow to run FCI using dodiscover? I am finding some of the options a bit confusing.

jaron-lee:
There is a nice figure in the Zhang paper that gives an example of applying R5, in Figure 4, page 1881. It looks exactly like what you wrote Adam, I think you're right.

https://www.sciencedirect.com/science/article/pii/S0004370208001008

adam2392:
> @adam2392 is there an example online that I can follow to run FCI using dodiscover? I am finding some of the options a bit confusing.

It's hard to parse, so can you elaborate a bit on what's confusing? Though not the FCI algorithm, this is an example of the PC: https://www.pywhy.org/dodiscover/dev/auto_examples/plot_pc_alg.html#sphx-glr-auto-examples-plot-pc-alg-py.

It's on my TODO list to refactor the algorithm API tho.

adam2392:
I think possibly we are also missing a unit-test for testing the entire FCI algorithm in the presence of selection bias, so that's an issue to get raised in dodiscover. The unit-test I posed: https://github.com/py-why/pywhy-graphs/pull/93#issuecomment-1777823431 should be sufficient.

@aryan26roy if you would like, you can just finish this PR w/o worrying about selection bias, but open up a GH issue to track what was not done here? This would help move things along incrementally. Simultaneously, it would probably help  us if you give us some insight onto what is confusing regarding running FCI and open up a GH issue about that downstream in dodiscover. 

It would be nice if we can just finish this sequence of PRs to get #67 working.

aryan26roy:
@adam2392 So far I have tested multiple examples and it's just one code path that is not tested(which will be tested only if I have a PAG with a --o edge). Everything else is tested and in my opinion, complete. I don't think it's a bad idea to finish this PR with a GH issue stating what is required for that last test.

aryan26roy:
@adam2392 I have added a GH issue, is this ok to merge?

aryan26roy:
> Meanwhile, there's still some CI errors, so will need to resolve those.

The errors are for networkx : ERROR: Package 'networkx' requires a different Python: 3.9.13 not in '>=3.10'

pywhy_graphs/algorithms/pag.py
```diff
@@ -908,3 +909,269 @@ def _check_ts_node(node):
...
+        if len(candidate_k) > 0:
+            # Make i-j into i->j
+            # logger.info(f"R2: Removing edge {i}-{j} to form {i}->{j}.")
+            graph.orient_uncertain_edge(i, j)
+            added_arrows = True
+    return added_arrows
+
+
+def pag_to_mag(graph):
+    """Convert a PAG to an MAG using algorithm defined in Theorem 2 of
```
adam2392, line 1109:
```suggestion
    """Sample a MAG from a PAG using Zhang's algorithm.
    
         Using algorithm defined in Theorem 2 of
```

pywhy_graphs/algorithms/pag.py
```diff
@@ -908,3 +909,269 @@ def _check_ts_node(node):
...
+     defined in :footcite:`Zhang2008`.
+     The algorithm turns all o-> edges to -> and -o edges to ->. Then converts the
+     input graph into a DAG with no unshielded colliders.
```
adam2392, line 1112:
```suggestion
     :footcite:`Zhang2008`, turns all o-> edges to -> and -o edges to ->. Then converts the
     input graph into a DAG with no unshielded colliders.
```

adam2392 (changes requested):
Few nits:

The docstring needs to get cleaned up a bit. The unit-tests could also be better documented. It's hard to parse from an outside perspective what the test is doing

aryan26roy:
I added more context to the unit tests and cleaned up the docstring a bit. Do you want me to do something more?

adam2392 (approved):

adam2392:
Thanks @aryan26roy !

PR #112, opened 2024-06-08
[ENH] Add the ability to find Proper Possibly Directed Paths

### LINKED ISSUE #111: [ENH] Add the ability to find Proper Possibly Directed Paths in a Graph
This is in continuation of PR [#1148](https://github.com/py-why/dowhy/pull/1148) in Dowhy. First PR in an effort to provide seamless integration with Dowhy.

A couple of functions that can take a graphs and find Proper Possibly Directed Paths in a graph.

aryan26roy:
Add the ability to find Proper Possibly Directed Paths to allow integration with DoWhy.

Fixes #111 

- Add the necessary functions to find Proper Possibly Directed Paths in a Graph.

## Before submitting

## After submitting


adam2392:
The CIs should now work. You can also test locally ofc with pytest.

aryan26roy:
@adam2392 I have completed the implementation and it seems to be correct. However, there is one problem. The output is a list of dictionaries. And the dictionaries can be in any order inside the list which is why the test may fail right now. How do I compare them in an order agnostic way?

adam2392:
Is there any inspiration from network's method for testing `all_simple_paths`?

https://github.com/networkx/networkx/blob/main/networkx/algorithms/tests/test_simple_paths.py

aryan26roy:
They are returning a set of tuples. I am assuming the order of tuples inside a set won't matter during comparison because of it's hashibality. However, I cannot store dictionaries inside sets. Either I change my implementation to store paths in sets(tuple()) or I can do the tests using the brute force method I described above.

What do you want me to do?

adam2392:
Is there any technical reason to do dictionaries instead of triples that I'm missing?

pywhy_graphs/algorithms/generic.py
```diff
@@ -855,3 +856,99 @@ def all_vstructures(G: nx.DiGraph, as_edges: bool = False):
                 else:
                     vstructs.add((p1, node, p2))  # type: ignore
     return vstructs
+
+
+def check_back_arrow(G: ADMG, X, Y: set):
```
adam2392, line 861: Let's make this a private function?

Also, document that this checks if any node in `Y` points back to `X`, where `X` is a node, not a set? Or did I misunderstand that?

pywhy_graphs/algorithms/generic.py
```diff
@@ -855,3 +856,99 @@ def all_vstructures(G: nx.DiGraph, as_edges: bool = False):
...
+            temp_paths = paths[i].copy()
+            temp_paths[len(temp_paths)] = elem
+            temp_arr.append(temp_paths)
+
+        new_paths.extend(recursively_find_pd_paths(G, X, temp_arr, Y))
+
+    return new_paths
+
+
+def possibly_directed_path(G, X: Optional[Set] = None, Y: Optional[Set] = None):
```
adam2392, line 937: Can we add a docstring?

pywhy_graphs/algorithms/generic.py
```diff
@@ -855,3 +856,99 @@ def all_vstructures(G: nx.DiGraph, as_edges: bool = False):
...
+    else:
+        nbr_temp = G.neighbors(X)
+        nbr_possible = check_back_arrow(nbr_temp)
+        x_neighbors = []
+
+        for elem in nbr_possible:
+            temp = dict()
+            temp[0] = X
+            temp[1] = elem
+            x_neighbors.append(temp)
```
adam2392, line 950: I'm confused why there's two paths depending on the type of X. Can't we just use `{G.neighbors(x) for x in X}` to get a set of all neighbors of X if X is a set?

adam2392 (commented):
Just left a few comments to hopefully push in the right direction

aryan26roy:
Sorry for the late reply.

> Is there any technical reason to do dictionaries instead of triples that I'm missing?

Not really, I just did not think of using tuples at the time. I think it would be easy to go from List<Dict> to Set<Tuples>. A better approach in terms of memory as well.

pywhy_graphs/algorithms/generic.py
aryan26roy: Was about to do it. I just wanted to know your opinion on how to test it, the PR is not ready for a full review yet. I should have been more explicit. 😅

pywhy_graphs/algorithms/generic.py
aryan26roy: Yep!

pywhy_graphs/algorithms/generic.py
aryan26roy: I did not know that G.neighbors(x) can take a set as an argument as well. Thanks for pointing this out.

aryan26roy: I tried running G.neighbors on a set of nodes and it errors out:

`TypeError: unhashable type: 'set'`

Is it supposed to throw this error or is it a bug?

pywhy_graphs/algorithms/generic.py
aryan26roy: > Also, document that this checks if any node in `Y` points back to `X`, where `X` is a node, not a set? Or did I misunderstand that?

Yes you understood that correctly, it is never called for a set X.

pywhy_graphs/algorithms/generic.py
adam2392: Ah yeah that is correct. I meant can we just have a one liner that iterates over the nodes in X and adds to a nbr set? The resulting nbr set is just all the nodes that are nbrs of X.

Or is there anything more nuanced happening?

aryan26roy:
@adam2392 I just realised, since in this implementation, all the first nodes of all the paths are in X, this finds all the proper possibly directed paths, doesn't it?

pywhy_graphs/algorithms/generic.py
aryan26roy: I wrote that function to account for the fact that the neighbor of a member of X might be another member of X. Given the following definition of a proper path:
>  A path from X to Y is proper (wrt X) if only its first node is in X

Am I wrong? Or do we need to to do this check?

adam2392:
Perhaps it's easier to implement just the function wrt a single node rather than a set of nodes. Do we need the functionality with X as a set? Wdyt?

aryan26roy:
The paper assumes everywhere that X will be a set. And the current implementation already covers and tests X being a set. It seems to me that changing the implementation now to only support a single node in X would be wasted effort.

adam2392:
> @adam2392 I just realised, since in this implementation, all the first nodes of all the paths are in X, this finds all the proper possibly directed paths, doesn't it?

Yes I believe so.

aryan26roy:
@adam2392 I believe the PR is complete. Can you do one last round of review? (PS. The failing test is due to some environment issue with some existing test)

pywhy_graphs/algorithms/generic.py
```diff
@@ -855,3 +856,161 @@ def all_vstructures(G: nx.DiGraph, as_edges: bool = False):
...
+    """Find all the possibly directed paths in a graph.
+
+    Parameters
+    ----------
+    G : DiGraph
+        A directed graph.
+    X : Set
+        Source.
+    Y : Set
+        Destination
```
adam2392, line 992: What if both are None? I don't think None should even be allowed, but lmk if I misunderstood

pywhy_graphs/algorithms/generic.py
```diff
@@ -855,3 +856,161 @@ def all_vstructures(G: nx.DiGraph, as_edges: bool = False):
...
+    for elem in Y:
+        if not (
+            G.has_edge(X, elem, G.bidirected_edge_name) or G.has_edge(elem, X, G.directed_edge_name)
+        ):
+            out.update(elem)
+
+    return out
+
+
+def _get_X_neighbors(G, X: set):
```
adam2392, line 890:
```suggestion
def _get_neighbors_of_set(G, X: set):
```

to be more explicit?

pywhy_graphs/algorithms/generic.py
```diff
@@ -855,3 +856,161 @@ def all_vstructures(G: nx.DiGraph, as_edges: bool = False):
...
+    """Retrieve all the neigbors of X when X has more than one element.
+
```
adam2392, line 892: You can add a note here saying if `X` is not a set, then one can simply use `graph.neighbors(X)`

adam2392 (commented):
Few short comments, but otw, it's looking good

pywhy_graphs/algorithms/tests/test_generic.py
```diff
@@ -496,3 +496,179 @@ def test_all_vstructures():
     # Assert that the returned values are as expected
     assert len(v_structs_edges) == 0
     assert len(v_structs_tuples) == 0
+
+
+def test_proper_possibly_directed():
```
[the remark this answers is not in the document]
aryan26roy: Separate them in separate functions or through comments?

pywhy_graphs/algorithms/generic.py
```diff
@@ -855,3 +856,161 @@ def all_vstructures(G: nx.DiGraph, as_edges: bool = False):
...
+            temp_paths = elem
+            temp_paths = temp_paths + (nbr,)
+            temp_set.add(temp_paths)
+
+        new_paths.update(_recursively_find_pd_paths(G, X, temp_set, Y))
+
+    return new_paths
+
+
+def proper_possibly_directed_path(G, X: Optional[Set] = None, Y: Optional[Set] = None):
```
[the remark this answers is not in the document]
aryan26roy: Got it!

pywhy_graphs/algorithms/generic.py
aryan26roy: Yes you are right, neither can ever be None. Maybe I should return an empty set whenever either one of them is None.

pywhy_graphs/algorithms/generic.py
aryan26roy: Sure.

pywhy_graphs/algorithms/generic.py
aryan26roy: Alright.

pywhy_graphs/algorithms/tests/test_generic.py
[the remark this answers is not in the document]
adam2392: Functions

pywhy_graphs/algorithms/generic.py
adam2392: Why not just not have default argument?

aryan26roy:
@adam2392 I have incorporated all of your suggestions. Can you take another look?

adam2392 (approved):
LGTM except two minor nits

pywhy_graphs/algorithms/generic.py
```diff
@@ -855,3 +856,187 @@ def all_vstructures(G: nx.DiGraph, as_edges: bool = False):
...
+    graph.
+
+    Parameters
+    ----------
+    G : DiGraph
+        A directed graph.
+    X : Set
+        Source.
+    Y : Set
+        Destination
```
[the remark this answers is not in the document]
aryan26roy: Oh yeah!

pywhy_graphs/algorithms/generic.py
```diff
@@ -855,3 +856,187 @@ def all_vstructures(G: nx.DiGraph, as_edges: bool = False):
...
+        x_neighbors = []
+
+        for elem in nbr_possible:
+            temp = dict()
+            temp[0] = X
+            temp[1] = elem
+            x_neighbors.append(temp)
+
+    path_list = _recursively_find_pd_paths(G, X, x_neighbors, Y)
+    print(path_list)
```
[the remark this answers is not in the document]
aryan26roy: Removed.

aryan26roy:
Done!

aryan26roy:
@adam2392 This can be optimised further by switching to a stack based implementation rather than recursion. We can add this to the TODO list.

adam2392:
Thanks @aryan26roy !

PR #119, opened 2024-08-26
[ENH] Add the ability to check if an edge is "visible"

### LINKED ISSUE #115: FEA A function to determine if an edge is "visible"
In causal graphs (DAG, ADMG, CPDAG, PAG, MAG), there is the notion of an edge being visible. For the DAG/ADMG/CPDAG, edges are by definition always visible.

Implement function `visible_edge(graph, source, target)` to determine if the edge between source and target is visible or not.

One can follow the papers https://auai.org/uai2015/proceedings/papers/155.pdf and https://www.jmlr.org/papers/volume18/16-319/16-319.pdf, and figures to use as unit-test cases.

**Additional context**
https://github.com/py-why/dowhy/issues/402#issuecomment-1404457573

aryan26roy:
Fixes #115 

- Add a function to check if the given edges in a path are "visible"

## Before submitting

## After submitting


aryan26roy:
@adam2392 What does this sign mean? 
>V •→X

Does it mean that the mark at the V end can be anything or that this can be a path of arbitrary length with the last mark being an arrow pointing towards X?

adam2392:
That is answered on pg2 of https://auai.org/uai2015/proceedings/papers/155.pdf

aryan26roy:
Does a function already exists to find if a collider path exists between two nodes in a graph?

adam2392:
> Does a function already exists to find if a collider path exists between two nodes in a graph?

I don't think so, but `all_vstructures` is similar, so you can modify that implementation to get a new function `all_collider_paths(src, target, G)`.

aryan26roy:
@adam2392 The paper says:

>A path of length one is a trivial collider path.

Do they mean a triple? (Only counting the non-endpoint node for the length) Because I don't see how a path could be a collider path otherwise.

adam2392:
> @adam2392 The paper says:
> 
> > A path of length one is a trivial collider path.
> 
> Do they mean a triple? (Only counting the non-endpoint node for the length) Because I don't see how a path could be a collider path otherwise.

I think they just mean a path of length one. So a single variable path is a collider, but yeah you would really only look at the path between two points usually. 

So a collider path is a sequence of variables <X_1, X_2, X_3, …, X_i>, where you just look at the edges between each consecutive triple <X_i-1, X_i, X_i+1> in that path to determine if they meet the collider condition.

pywhy_graphs/algorithms/tests/test_generic.py
```diff
@@ -496,3 +496,32 @@ def test_all_vstructures():
...
+
+    assert True == check_visibility(pag, "X", "Y")
+
+    pag = PAG()
+    pag.add_edge("Y", "X", pag.directed_edge_name)
+    pag.add_edge("Z", "Y", pag.bidirected_edge_name)
+    pag.add_edge("Z", "K", pag.bidirected_edge_name)
+    pag.add_edge("K", "H", pag.bidirected_edge_name)
+
+    assert False == check_visibility(pag, "X", "Y")
```
aryan26roy, line 526: @adam2392 Just want to check with you, this case should be false, right?

